PR #14, opened 2015-10-14
Godeps: remove wrongly imported binary.

sgotti:


sgotti:
Already fixed by sorintlab/pollon#3 and #15 .

PR #98, opened 2015-12-03
Support pg_rewind

s-hironobu:
This is the first implementation to use pg_rewind. If pg_rewind can be used, stolon uses it to synchronize a standby with the primary.

This implementation can be run on a simple cluster. For using pg_rewind, assume that username is "postgres" and password is none.

### DIFF
--- a/cmd/keeper/keeper.go
+++ b/cmd/keeper/keeper.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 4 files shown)


sgotti:
@s-hironobu Thanks a lot for your PR! I'll take a few days to try it and also see how it can be tested (need to install pg_rewind on pg 9.4 or update to pg 9.5 on semaphore).

In addition it'll be good to detect if pg_rewind cannot sync due to missing WALs and fallback to pg_basebackup.

s-hironobu:
I was thinking version 9.5 is released within this year, but I guessed wrong. It seems that there's no need to hurry.

This patch is a feasibility study for the pg_rewind, so I'm glad to be confirmed that pg_rewind can be run on Stolon without major changes. I think superuser's password management and user authentication may be problems when using this in practical systems. pg_rewind should be accessed as superuser from an other host/container.


BTW, stolon's proxy can be added DSR(direct server return) or Virtual IP feature? Proxy might will be performance bottleneck, as well as security hole, but if DSR or Virtual IP can be used, Stolon can avoid those problems. (Pacemaker supports VIP.)

Best regards,

shaneog:
With 9.5 released since January, is there a way `pg_rewind` support could be added?

sgotti:
@shaneog Sure, starting from the @s-hironobu work and suggestions we have some initial work on test cases. We need to determine when pg_rewind cannot be used and fallback to a pg_basebackup. We'd like to update/open a PR in the next days/week.

Do you have some special needs for pg_rewind (for example big databases) that are creating problems with stolon using only pg_basebackup?

shaneog:
@sgotti I'm just about to test stolon (and patroni) and have a mid-sized database to use. I will see how it goes during testing, but at the moment I don't have any examples of specific problems.

sgotti:
@shaneog ok good to know. As a note, we just created stolon because other solutions are less resilent to network partitioning and we prefer consistency over availability. Let me know if you need any help.

sgotti:
@s-hironobu Thanks for your work and suggestions. In the past days we did a lot of heavy lifting to better integrate pg_rewind. I opened #122 which adds support for pg_rewind, tests and other goodies (it can also work with synchronous replication).

@shaneog @s-hironobu If you're interested please take a look at it.

sgotti:
#122 has been merged.

PR #122, opened 2016-03-17
keeper: add pg_rewind support

sgotti:
This patch adds pg_rewind support. By default it's disabled and can be
enabled setting the `use_pg_rewind` cluster config options to true.
Since pg_rewind needs to connect to the master instance with a superuser
role, the superuser credential needs to be provided (currently at
startup using environment variables or options (discouraged)) to all the
`stolon-keeper` processes. For security reason they cannot be provided
inside the cluster config.

It'll also work with synchronous replication (pg_rewind needs to connect
with `synchronous_commit` disabled at its session level).

Add related tests.

Add related documentation.


sgotti:
@s-hironobu @shaneog Merging this (it's working well and can be changed/improved in future). If you're going to test pg_rewind support please report issues or feature requests. Thanks.

PR #130, opened 2016-03-30
In the proxy example, the environment variables for the storage backe…

robdaemon:
…nd are incorrect (copy/paste error?)


sgotti:
Good catch! That was introduced by the migration to libkv (cbe0782446f00d68c0a806f1613c703647986e86)

This makes me think we'll need to find a way to test also on a kubernetes cluster (I'll check if it willl be possible to setup a simple kubernetes cluster on a semaphore ci vm).

PR #133, opened 2016-04-04
Setting the initial pgsql superuser password

robdaemon:
* Adds initial-su-username and initial-su-password-secret-file parameters. If these are set, stolon will set the proper password for the initial superuser when postgresql is first initialized.
* Adds an example secret and usage to the Kubernetes examples.
* Adds support for specifying a namespace to sentinel. This allows stolon to be ran outside of the "default" namespace.

### DIFF
--- a/cmd/keeper/keeper.go
+++ b/cmd/keeper/keeper.go
@@ -57,20 +57,22 @@ var cmdKeeper = &cobra.Command{
 }
 
 type config struct {
-	id              string
-	storeBackend    string
-	storeEndpoints  string
-	dataDir         string
-	clusterName     string
-	listenAddress   string
-	port            string
-	pgListenAddress string
-	pgPort          string
-	pgBinPath       string
-	pgConfDir       string
-	pgSUUsername    string
-	pgSUPassword    string
-	debug           bool
+	id                      string
+	storeBackend            string
+	storeEndpoints          string
+	dataDir                 string
+	clusterName             string
+	listenAddress           string
+	port                    string
+	pgListenAddress         string
+	pgPort                  string
+	pgBinPath               string
+	pgConfDir               string
+	pgSUUsername            string
+	pgSUPassword            string
+	debug                   bool
+	pgInitialSUUsername     string
+	pgInitialSUPasswordFile string
 }
 
 var cfg config
@@ -93,6 +95,8 @@ func init() {
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgConfDir, "pg-conf-dir", "", "absolute path to user provided postgres configuration. If empty a default dir under $dataDir/postgres/conf.d will be used")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUUsername, "pg-su-username", "", "postgres superuser user name (required by pg_rewind)")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPassword, "pg-su-password", "", "postgres superuser password (required by pg_rewind)")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.pgInitialSUUsername, "initial-pg-su-username", "", "postgres initial superuser username")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.pgInitialSUPasswordFile, "initial-pg-su-password-file", "", "postgres initial superuser password secret file")
 	cmdKeeper.PersistentFlags().BoolVar(&cfg.debug, "debug", false, "enable debug logging")
 }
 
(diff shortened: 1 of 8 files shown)


cmd/keeper/keeper.go
```diff
@@ -93,6 +95,8 @@ func init() {
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgConfDir, "pg-conf-dir", "", "absolute path to user provided postgres configuration. If empty a default dir under $dataDir/postgres/conf.d will be used")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUUsername, "pg-su-username", "", "postgres superuser user name (required by pg_rewind)")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPassword, "pg-su-password", "", "postgres superuser password (required by pg_rewind)")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.initialSUUsername, "initial-su-username", "", "postgres initial superuser username")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.initialSUPasswordFile, "initial-su-password-secret-file", "", "postgres initial superuser password secret file")
```
sgotti: I also wanted to read it from a file when adding `pgSUPassword` for pg rewind instead of relying of an argument or env variable. I'll create a future PR the make it consistent with your proposal.

cmd/keeper/keeper.go
```diff
@@ -113,9 +131,6 @@ func (p *PostgresKeeper) getSUConnParams(keeperState *cluster.KeeperState, setPa
 		"dbname":           "postgres",
 		"sslmode":          "disable",
 	}
-	if setPassword {
-		cp["password"] = p.cfg.pgSUPassword
-	}
```
sgotti: Not sure why I addedt the setPassword bool but it's actually not needed... thanks.

cmd/keeper/keeper.go
sgotti: To make it consistent I'd like to leave the pg prefix. But the options are becoming quite long :stuck_out_tongue_closed_eyes: 

what about:

`initial-pg-su-username`
`initial-pg-su-password-file`
?

cmd/keeper/keeper.go
```diff
@@ -103,7 +107,21 @@ var defaultPGParameters = pg.Parameters{
 	"hot_standby":             "on",
 }
 
-func (p *PostgresKeeper) getSUConnParams(keeperState *cluster.KeeperState, setPassword bool) pg.ConnParams {
+func (p *PostgresKeeper) getInitialSUPassword() string {
+	var pw string
+
+	if p.cfg.initialSUPasswordFile != "" {
+		pwBytes, err := ioutil.ReadFile(p.cfg.initialSUPasswordFile)
```
sgotti: should we check file permissions to be <= than 600. Perhaps that's a lesser problem inside a pod but on a standard server it'll be good to make them readable only by the stolon's owner.

But looks like that kubernetes actually doesn't implement setting file permission so actually it should ignore the check on kubernetes. Perhaps just add a TODO at the moment...
Haven't tried so I'm not sure it will be possible to them with that user / permissions using kubernetes secrets?

cmd/keeper/keeper.go
```diff
@@ -103,7 +107,21 @@ var defaultPGParameters = pg.Parameters{
...
+		if err != nil {
+			log.Errorf("Unable to read password from file %s, error: %s", p.cfg.initialSUPasswordFile, err.Error())
```
sgotti: `log.Errorf("Unable to read password from file %s, error: %v", p.cfg.initialSUPasswordFile, err)` to make it consistent with other errors logging.

sgotti:
@robdaemon thanks a lot for this patch, that's really cool. I put some comments inline.

robdaemon:
Give me an hour or so, I'll push up a commit that addresses your comments!

robdaemon:
@sgotti okay, addressed your feedback. I added a TODO about enforcing the file permissions, and a log warning if they're too open. I chose a warning on the logging because it's a recoverable condition.

sgotti:
@robdaemon Great! Thanks a lot! I'm going to merge it.

sgotti:
@robdaemon I forgot to ask if you can please squash the commits in a single one.

robdaemon:
done! I usually hold off on squashing until it meets upstream approval :)

sgotti:
> done! I usually hold off on squashing until it meets upstream approval :)

You're right, I was just too eager to merge it.

Thanks again!

PR #140, opened 2016-06-29
Update vendoring

agalitsyn:
What was done:
* Upgrade github.com/docker/libkv to latest version to avoid some fixed bugs
* Upgrade vendoring scheme to use normal import paths and `GO15VENDOREXPERIMENT`


agalitsyn:
I close in in favor of #141, because it's easier to merge, only vendor scheme changed, libs are the same version

PR #141, opened 2016-06-30
Godep workspace -> vendor/

agalitsyn:
* Upgrade vendoring scheme to use standard import paths and GO15VENDOREXPERIMENT


sgotti:
@agalitsyn Some months ago I open but never merged #131. I opted for glide + glide-vc since it looked like a cleaner way to handle dependencies and pin their versions without messing with the GOPATH or creating a clean GOPATH with the wanted version for every dependency. Now the same idea is used in other projects like CoreOS rkt and other.

What're your thoughts on this? Will you mind to rebase your work starting with #131?

agalitsyn:
from my point of view there is no dramatic difference between glide and godeps, both do vendoring, so if you prefer glide we can close PR

sgotti:
@agalitsyn Thanks, I'll go ahead with glide and update #131 (also updating libkv as in #140).

sgotti:
Done in #131.

PR #142, opened 2016-06-30
Godep workspace -> vendor/

agalitsyn:
* Upgrade vendoring scheme to use standard import paths and GO15VENDOREXPERIMENT

Related to https://github.com/sorintlab/stolon/pull/141, we can add it to master earlier than in upstream


agalitsyn:
dup of #141

PR #147, opened 2016-09-06
Restrict display of replication password

wchrisjohnson:
Restrict the display of the replication password to debug logging level. Keep the rest of the existing log.infof information.


sgotti:
@wchrisjohnson LGTM. Thanks!

PR #149, opened 2016-09-14
Changed to official PostgreSQL docker image

tomazzaman:
Changes:
- removed the image in kubernetes example because it had no control over postgres version (or any other version for that matter)
- `run.sh` is too simple to bother having around, rather moved those to YAML files (also having env variable dictate which command to run _feels_ wrong)
- moved the `Dockerfile` to the root so that Stolon is built _inside_ the container, rather than outside, which can have a different environment
- said `Dockerfile` now uses the official image


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    && rm -rf /var/lib/apt/lists/* \
+    && curl $GO_DL_URL | tar xvz -C /usr/local
+
+COPY . /usr/src/stolon
+
+RUN cd /usr/src/stolon \
+    && ./build \
+    && mv bin/* /usr/local/bin/
+
+COPY run.sh /usr/local/bin
```
thpham: I got an error here, because run.sh is not at the root path. with your approach you can even remove it.

build
```diff
@@ -32,9 +32,9 @@ export GO15VENDOREXPERIMENT=1
 
 use_go_install=
 
-go_root_dir=$(go env GOROOT)
-go_host_os=$(go env GOHOSTOS)
-go_host_arch=$(go env GOHOSTARCH)
+go_root_dir=$(go env GO_ROOT)
+go_host_os=$(go env GO_HOST_OS)
+go_host_arch=$(go env GO_HOST_ARCH)
```
sgotti: These will break the below `cgodisabled_pkg_dir` variable since `go env GO_ROOT` will return an empty string (GO_ROOT and the other are not valid go env variables).
I think there's no reason to change this.

examples/kubernetes/stolon-keeper.yaml
```diff
@@ -16,9 +16,16 @@ spec:
...
       - name: stolon-keeper
         image: sorintlab/stolon:master
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            export POD_IP=$(hostname -i)
+            export STKEEPER_LISTEN_ADDRESS=$POD_IP
+            export STKEEPER_PG_LISTEN_ADDRESS=$POD_IP
+            chown stolon:stolon $PGDATA
```
sgotti: I don't see where PGDATA is defined.

examples/kubernetes/stolon-keeper.yaml
```diff
@@ -16,9 +16,16 @@ spec:
...
+            exec gosu stolon stolon-keeper --data-dir $PGDATA
```
sgotti: Is `gosu` already available in the postgresql image?

examples/kubernetes/image/docker/run.sh
```diff
@@ -1,52 +0,0 @@
-#!/bin/bash
-
-function setup() {
-  # use hostname command to get our pod's ip until downward api are less racy (sometimes the podIP from downward api is empty)
-  export POD_IP=$(hostname -i)
-}
-
-function checkdata() {
-  if [[ ! -e /stolon-data ]]; then
```
sgotti: This check, warning and data creation were removed but it'll be worth keeping them. The warning is not completely correct since already having a /stolon-data doesn't mean that it'll be persisted (it should be mounted on an external volume and that volume should be a k8s persistent volume) but I'm not sure the check should go so deeper.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN groupadd -r stolon --gid=543 && useradd -r -g stolon --uid=543 stolon
+
+ENTRYPOINT [ "/usr/local/bin/run.sh" ]
```
sgotti: run.sh was removed, so the image doesn't have a default entrypoint, right?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ENV GO_DL_BASE https://storage.googleapis.com/golang
+ENV GO_DL_URL $GO_DL_BASE/go$GO_VERSION.$GO_HOST_OS-$GO_HOST_ARCH.tar.gz
+ENV PATH $GO_ROOT/bin:$PATH
+
+RUN set -x \
+    && apt-get update \
+    && apt-get install -y curl \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl $GO_DL_URL | tar xvz -C /usr/local
+
```
sgotti: To keep the image as small as possible I'll do some cleanup after the build to remove golang, apt data etc...

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM postgres:9.5.4
+
+ENV GO_VERSION 1.7.1
+ENV GO_HOST_ARCH amd64
+ENV GO_HOST_OS linux
+ENV GO_ROOT /usr/local/go
+ENV GO_DL_BASE https://storage.googleapis.com/golang
+ENV GO_DL_URL $GO_DL_BASE/go$GO_VERSION.$GO_HOST_OS-$GO_HOST_ARCH.tar.gz
```
sgotti: These env variables will be exported to the app but they are only needed to compose the `GO_DL_URL` (and are creating some confusion with the go ones). I'd avoid using env variables as shell variables just do something like:

```
curl  https://storage.googleapis.com/golang/go1.7.1.linux-amd64.tar.gz
```
since all that variables are fixed anyway.

sgotti:
@tomazzaman Thanks for this PR! I reviewed it but haven't tested it again.

> moved the Dockerfile to the root so that Stolon is built inside the container, rather than outside, which can have a different environment

The Dockerfile was meant just for the k8s example and not for an "official" stolon docker image so I'd like to keep it inside the example dir. That's the reason why the build script copied the binaries (since a symlink of the repository root inside `examples/kubernetes/image/docker` was, at least some months ago, ignored by docker).
I'm not sure which is the best way to build stolon inside docker while keeping the DockerFile in `examples/kubernetes/image/docker`.

I'm also not sure if just one image is the best approach. For example there's no need to have a postgresql image for stolon-sentinel and stolon-proxy but they can be (like done now) statically compiled and we can create a docker image with just that files.

Dockerfile
tomazzaman: Ah yes, I actually planned to, but forgot :/

examples/kubernetes/stolon-keeper.yaml
tomazzaman: Since it's moving to the official image, $PGDATA is defined in there.

examples/kubernetes/stolon-keeper.yaml
tomazzaman: Yes

Dockerfile
tomazzaman: True, forgot to remove.

tomazzaman:
@sgotti regarding the Dockerfile, there's unfortunately now way to copy contents inside the image from directories outside the one that Dockerfile is in, so it had to be moved to the root - no way around it, apparently for security reasons.

And regarding just one image, I actually built another image just for proxy and sentinel, but then decided to abandon that approach because if we have only one image (even if it's bigger because of Postgres) then the same image is used everywhere - resulting in Kubernetes being able to reuse it for different purposes rather than downloading a new, separate iamge. This I think adds to better UX.

Dockerfile
tomazzaman: Agreed, will fix.

sgotti:
> @sgotti regarding the Dockerfile, there's unfortunately now way to copy contents inside the image from directories outside the one that Dockerfile is in, so it had to be moved to the root - no way around it, apparently for security reasons.

Yeah, the build context sent to the docker daemon won't permit getting file outside the Dockerfile dir. I can see two ways:

* Keep it as you provided and make it something as an official stolon image.
* Use the current build file (or create a build file/Makefile inside `examples/kubernetes/image/docker`) that needs to be executed before the `docker build` that will create a shallow git clone of the repository root or just copies the required contents inside `examples/kubernetes/image/docker`. This is what is done now but will requires users running an additional/different command than just `docker build`.

> And regarding just one image, I actually built another image just for proxy and sentinel, but then decided to abandon that approach because if we have only one image (even if it's bigger because of Postgres) then the same image is used everywhere - resulting in Kubernetes being able to reuse it for different purposes rather than downloading a new, separate iamge. This I think adds to better UX.

I agree, the bigger image will be needed anyway to run the stolon-keeper and there won't be much overhead caused by unused data for the sentinel and proxies containers since the docker graphdrivers will make them share the same underlying image.

euank:
> regarding the Dockerfile, there's unfortunately now way to copy contents inside the image from directories outside the one that Dockerfile is in, so it had to be moved to the root - no way around it, apparently for security reasons.

The build context directory doesn't have to match the Dockerfile directory. It could remain where it is now and the command could be changed to `docker build -f ./examples/kubernetes/image/docker/Dockerfile -t image-name .` and be run with its working directory at the root, and it should work!

sgotti:
I opened #201 that introduces the use of postgres official images and a Makefile to build them starting from a template Dockerfile where you can define the postgres image version to use, in this way there's no need to have multiple duplicated files. The run.sh removal and the build inside docker has been left out and can be implemented in future PRs or rebasing this PR on it.

sgotti:
Closing since we are now using postgresql "official" image. The build inside the image, if really wanted, could be done in another dedicated PR.

PR #158, opened 2016-10-10
Reproduce readlink -f behavior on osx

killix:
The script failed to run on OSx because `readlink -f` is not supported by Darwin system.

### DIFF
--- a/build
+++ b/build
@@ -2,7 +2,28 @@
 
 set -e
 
-BASEDIR=$(readlink -f $(dirname $0))
+# Cross compatibility with osx
+# origin source: https://github.com/kubernetes/kubernetes/blob/master/hack/lib/init.sh#L102
+function readlinkdashf() {
+	# run in a subshell for simpler 'cd'
+  (
+    if [[ -d "$1" ]]; then # This also catch symlinks to dirs.
+      cd "$1"
+      pwd -P
+    else
+      cd $(dirname "$1")
+      local f
+      f=$(basename "$1")
+      if [[ -L "$f" ]]; then
+        readlink "$f"
+      else
+        echo "$(pwd -P)/${f}"
+      fi
+    fi
+  )
+}
+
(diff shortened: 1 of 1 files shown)


sgotti:
@killix Thanks for the PR! I'm not sure but perhaps `readlink -f` can be replaced by something like `${BASH_SOURCE[0]` (haven't tried, I just remember I did this in other scripts).

Out of curiosity, are you using stolon on MacOS? Perhaps we can at least test builds enabling macos on Travis.

killix:
``dirname "${BASH_SOURCE[0]}"`` will return the relative path ``.`` and not the fulle path.
I am only using Stolon in my MacOs for development purpose and when a i not can/want tu launch docker.

build
```diff
@@ -2,6 +2,31 @@
 
 set -e
 
+# Cross compatibility with osx
+# http://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac
```
sgotti: Just noticed a simpler version used by kubernetes here: https://github.com/kubernetes/kubernetes/blob/master/hack/lib/init.sh#L99
What do you think?

build
```diff
@@ -2,6 +2,31 @@
...
+if [ "$(uname)" == "Darwin" ]; then
+	function readlink() {
```
sgotti: Just for clearness I'll call this `readlinkdashf` or something similar.

killix: 👍

killix:
sorry for the time. PR updated.

sgotti:
@killix great! Can you please rebase on the current master? (I just triggered the github button to merge the master into your PR to see if the tests passed but I'd like to avoid this and have a clean merge...)

killix:
done.

sgotti:
LGTM! Thanks again!

Issue #144, opened 2016-07-29
Feature Request: Point in Time Recovery (PITR)


bdurrow:
I am interested in bringing up production postgres databases in OpenShift Origin.  It looks like there has been some work to template stolon for OpenShift so it might be a good fit.  Part of our requirements include Point In Time Recovery (PITR) but I don't see any backup/restore features in stolon.  Is it on the roadmap?

I have done some work with pghoard (https://github.com/ohmu/pghoard) to have it work in a docker and plan to do more work to allow it to serve archived WALs for recovery out of a docker (with object store backing).  Does this fit architecturally with Stolon?

wied03:
It doesn't fix this, but I had a rough procedure I used to use with corosync/pacemaker for this that worked. I used WAL-E for backup so this kind of depended on that. Maybe this might help the process though.

The rough steps were:
1. Shutdown the corosync cluster
2. Wipe out the PGSQL data directory
3. Setup a recovery.conf as noted below
4. Run wal-e backup-fetch
5. Start just a single node (the master) without the corosync/pacemaker stuff (the standard pgsql way)
6. Verify the restore looks good on the master
7. Run ```select pg_xlog_replay_resume();``` on the master
8. Wait for ```select pg_is_in_recovery();`` to indicate recovery is over
9. Once recovery is over, shutdown the standalone instance and start the master via Pacemaker (but with slaves turned off)
10. Once the master comes up OK, Run pg_basebackup on the slaves to get a complete baseline set of data again
11. Cleanup the Pacemaker resource state
12. Re-enable the slaves

Recovery.conf:
```bash
 restore_command='envdir /etc/wal-e.d/env /usr/bin/wal-e wal-fetch %f %p'
 recovery_target_name = '<%=restore_point_name%>'
 # If we don't include latest, the default is the timeline of the base backup, which in a re-restore situation, won't work since we will have created a new timeline by then
 recovery_target_timeline = 'latest'
 ```

sgotti:
@bdurrow sorry for the late reply.

> I am interested in bringing up production postgres databases in OpenShift Origin. It looks like there has been some work to template stolon for OpenShift so it might be a good fit. Part of our requirements include Point In Time Recovery (PITR) but I don't see any backup/restore features in stolon. Is it on the roadmap?

There's some work going on backup feature.

About the restore feature I'd like to know some details on your use case.

Are you interested on restoring a db on a running stolon cluster or just restoring at cluster initialization?

[...]

PR #164, opened 2016-10-18
Update FAQ with partition/backup questions

wied03:
This is most certainly a draft but thought I'd get the ball rolling since others may have these questions.

Note the bolded text under 'Pure Kubernetes'. I suspect the shared volume is really the only reason. Not trying to be too negative on Stolon there. Just trying to understand this more myself and clearly explain the value add.

A lot of this dialog was straight from our gitter talk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,30 @@ For our need to forcibly close connections to unelected masters and handle keepe
 
 We are open to alternative solutions (PRs are welcome) like using haproxy if they can met the above requirements. For example, an hypothetical haproxy based proxy needs a way to work with changing ip addresses, get the current cluster information and being able to forcibly close a connection when an haproxy backend is marked as failed (as a note, to achieve the latter, a possible solution that needs testing will be to use the [on-marked-down shutdown-sessions](https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.2-on-marked-down) haproxy server option).
 
+### How does Stolon differ from ____?
+
+#### Pure Kubernetes
+
+Stolon does not require a shared volume be attached to whatever node the Postgres container is running on. **Is Stolon faster than K8's pod process?? Is that another reason? I can't imagine it would be that different**. With a pure Kubernetes approach and something like an EBS volume, you can have a single Postgres master and even slave nodes with replication that could serve as read only servers. In the event of a master failure, K8S can move the DB container around and repoint the service as necessary.
+
+#### Pacemaker
+
+Stolon eliminates the fencing/shared storage requirement because:
+1. It can rely on etcd/consul clustering in the event of partitions
+2. It makes more assumptions (consistency) about what to do in these events (see the `testTimelineFork` integration test for some examples.)
+
+### Why is fencing not necessary with Stolon?
+
+First, no shared storage. Then we can have multiple cases of partitioning. So we try to solve this primarily in two ways: 1) The sentinel computes a "wanted" component states (now called cluster view). The keepers and the proxies try to converge to this states. 2) The stolon-proxy is the one that forces clients to the right master and forcibly closes connection to unelected ones. So you have to connect to the master only via the proxy. So, if all the stolon components behave correctly we can avoid fencing.
+
+### What happens if etcd is partitioned?
+
+Etcd isn't a problem. Due to Raft it'll only work with a quorum so there's no way to write to partition without quorum. In this case all the stolon will get an error and then retry in the next interval. Additionally the stolon-proxy, if it cannot talk with etcd, by default will drop all connection to the master since it cannot know if the cluster view has changed (for example if the proxy has problems talking with etcd but the sentinel can talk).
+
+### How are backups handled with Stolon?
+
+**TBD. Explain a little about WAL-E, how to configure Stolon to work with it and what happens when partitions, etc. occur.**
+
 ## Contributing to stolon
 
 stolon is an open source project under the Apache 2.0 license, and contributions are gladly welcomed!


README.md
```diff
@@ -75,6 +75,30 @@ For our need to forcibly close connections to unelected masters and handle keepe
 
 We are open to alternative solutions (PRs are welcome) like using haproxy if they can met the above requirements. For example, an hypothetical haproxy based proxy needs a way to work with changing ip addresses, get the current cluster information and being able to forcibly close a connection when an haproxy backend is marked as failed (as a note, to achieve the latter, a possible solution that needs testing will be to use the [on-marked-down shutdown-sessions](https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.2-on-marked-down) haproxy server option).
 
+### How does Stolon differ from ____?
+
+#### Pure Kubernetes
+
+Stolon does not require a shared volume be attached to whatever node the Postgres container is running on. **Is Stolon faster than K8's pod process?? Is that another reason? I can't imagine it would be that different**. With a pure Kubernetes approach and something like an EBS volume, you can have a single Postgres master and even slave nodes with replication that could serve as read only servers. In the event of a master failure, K8S can move the DB container around and repoint the service as necessary.
```
sgotti: > **Is Stolon faster than K8's pod process?? Is that another reason? I can't imagine it would be that different**

Yeah this point needs to be expanded: our original thoughts (not sure if all are valid anymore):
* k8s failed node detection/pod eviction took minutes.
* On aws an ebs volume is tied to an availability zone so you cannot failover to another availability zone. With stolon you can

README.md
```diff
@@ -75,6 +75,30 @@ For our need to forcibly close connections to unelected masters and handle keepe
...
+
+#### Pacemaker
+
+Stolon eliminates the fencing/shared storage requirement because:
+1. It can rely on etcd/consul clustering in the event of partitions
+2. It makes more assumptions (consistency) about what to do in these events (see the `testTimelineFork` integration test for some examples.)
```
sgotti: I'll add that it uses a proxy to redirect to the right master and forcibly close connections to unelected masters.

README.md
sgotti: Around k8s and fencing I'll add that a funny point is that I stated that k8s needs fencing some time ago (https://groups.google.com/forum/#!topic/coreos-dev/ZCAywIDVyys ignore the first message...) and it was converted to a k8s issue https://github.com/kubernetes/kubernetes/issues/12338 . Now finally this has been taken care in https://github.com/kubernetes/kubernetes/pull/34160

README.md
```diff
@@ -75,6 +75,30 @@ For our need to forcibly close connections to unelected masters and handle keepe
...
+
+### Why is fencing not necessary with Stolon?
+
+First, no shared storage. Then we can have multiple cases of partitioning. So we try to solve this primarily in two ways: 1) The sentinel computes a "wanted" component states (now called cluster view). The keepers and the proxies try to converge to this states. 2) The stolon-proxy is the one that forces clients to the right master and forcibly closes connection to unelected ones. So you have to connect to the master only via the proxy. So, if all the stolon components behave correctly we can avoid fencing.
+
+### What happens if etcd is partitioned?
```
sgotti: I'll change this to s/etcd/store/

README.md
```diff
@@ -75,6 +75,30 @@ For our need to forcibly close connections to unelected masters and handle keepe
...
+
+Etcd isn't a problem. Due to Raft it'll only work with a quorum so there's no way to write to partition without quorum. In this case all the stolon will get an error and then retry in the next interval. Additionally the stolon-proxy, if it cannot talk with etcd, by default will drop all connection to the master since it cannot know if the cluster view has changed (for example if the proxy has problems talking with etcd but the sentinel can talk).
```
sgotti: > Due to Raft it'll only work with a quorum so there's no way to write to partition without quorum. 

I'll add that we also use linearized reads so they'll be directed only to the partition with quorum.

README.md
```diff
@@ -75,6 +75,30 @@ For our need to forcibly close connections to unelected masters and handle keepe
...
+
+### How are backups handled with Stolon?
+
+**TBD. Explain a little about WAL-E, how to configure Stolon to work with it and what happens when partitions, etc. occur.**
+
```
sgotti: Actually I think this could be moved to another issue. I'm implementing #144 (but first it'll need additional changes) and I really want to add a wal-e example in it.

sgotti:
@wied03 Thanks for the PR! I just added some additional notes. Probably the form of the sentences could be improved but I'll try to do this in the next review pass (and I'm definetely not the right person to fix english sentences :stuck_out_tongue_closed_eyes: )

README.md
wied03: @sgotti - I'm guessing if you're using Stolon, because the sentinels' cluster view piece we talked about before, that Stolon isn't really affected one way or the other by the (current) lack of fencing in k8s?

wied03: On point 1, I haven't used it much yet so have no choice but to agree.
On point 2, I forgot about that. With Elastic File System, the different AWS offering, you can span multiple zones it looks like.

If that's true, then I suppose using k8s only with EFS can be fine if you can tolerate the k8s switchover time and/or pricing for EFS. If not, then Stolon is the place to be.

sgotti: I don't get your question... What does it mean "I'm guessing if you're using Stolon"?

> Stolon isn't really affected one way or the other by the (current) lack of fencing in k8s?

Lack of fencing in k8s depends on how you are setting up your pods. The different issues are stated in kubernetes/kubernetes#34160. So stolon is affected if you're encountering that issues like any other program running under k8s. 
For example in a scenario where you have N stolon-keeper pods manually defined with their PV (no petset) and X sentinels the lack of fencing will be an issue only if you're using iscsi or other form of pv that actually, due to the lack of fencing, can be mounted on two nodes causing data corruption. But we can't do anything to avoid this since the problem is in the underlying infrastructure (k8s) and needs to be fixed there. So you'll have to restore from backup or, with stolon you can detect this corruption's and just switch to an healthy standby.

wied03: @sgotti - I assumed that you would not use any persistent volumes (excluding emptyDir) with Stolon due to the shared nothing approach. If that's the case, then what sort of fencing would Stolon need to avoid problems?

sgotti: @wied03 Persistent volumes are definitely needed in production or you have to be sure that at least one pod with a synced stolon standby is always active.... But what will you do if you k8s cluster goes down? The same IMHO applies to other shared nothing DBs like cassandra, cockroachdb etc...

The k8s example is to just start trying it and not to be used in production (it's also explained here: https://github.com/sorintlab/stolon/tree/master/examples/kubernetes#create-the-first-stolon-keeper)

wied03: @sgotti - That's true, I didn't think about that and when I skimmed that page, I missed that remark. In any case, the dependency on that proposal for k8s being implemented might be good to illustrate. More importantly, what's the implication for Stolon until that's implemented? I think it would be good to illustrate the scenario so that people know the trade offs. What does "with stolon you can detect this corruption's and just switch to an healthy standby" mean in your statement above? Later I could go look in the code here to see if that is automatic or not but I do not have time right now.

sgotti:
After #219 I'll open a PR to update these FAQs and change them due to the new implemented features.

sgotti:
Reworked in #224.

Issue #168, opened 2016-10-20
Impossible to add a second node to the Stolon cluster [SOLVED]


afiskon:
Stolon version: 0.3.0 (same issue on master branch) - built from source
Consul version: 0.7.0 (default Arch Linux package)
PostgreSQL version: 9.5.4 (default Arch Linux package)
Go version: 1.7.1 (default Arch Linux package)

I've created 3 virtual machines using Virtual Box. VMs are in 10.0.3.0/24 network - 10.0.3.7 (archlinux1), 10.0.3.8 (archlinux2) and 10.0.3.9 (the last one is not used below).

Steps to reproduce.

On archlinux1:

```
./bin/stolon-sentinel --cluster-name mycluster --store-backend consul --listen-address 10.0.3.7

echo -n 'supasswd123' > ~/.pgsupasswd
echo -n 'replpasswd123' > ~/.pgreplpasswd

./bin/stolon-keeper --data-dir ./postgres1 --id postgres1 --cluster-name mycluster --pg-su-passwordfile ~/.pgsupasswd --pg-repl-username repluser --pg-repl-passwordfile ~/.pgreplpasswd --store-backend consul --pg-listen-address 10.0.3.7

./bin/stolon-proxy --cluster-name mycluster --listen-address 10.0.3.7 --port 25432 --store-backend consul

psql --host localhost --port 5432 -d postgres

CREATE DATABASE test;
CREATE USER test WITH PASSWORD 'qwerty';
GRANT ALL PRIVILEGES ON DATABASE test TO test;

psql --host 10.0.3.7 --port 25432 --user test
```

Single node configuration works as expected.

On archlinux2:

```
./bin/stolon-keeper --data-dir ./postgres2 --id postgres2 --cluster-name mycluster --pg-su-passwordfile ~/.pgsupasswd --pg-repl-username repluser --pg-repl-passwordfile ~/.pgreplpasswd --store-backend consul --pg-listen-address 10.0.3.8

./bin/stolon-sentinel --cluster-name mycluster --store-backend consul --listen-address 10.0.3.8
```

sentinel output:

```
2016-10-20 18:06:44.186737 [pkg_logger.go:134] I | sentinel: id: 96674841
2016-10-20 18:06:44.188273 [pkg_logger.go:134] I | sentinel: Trying to acquire sentinels leadership

(and nothing more)
```

keeper output:

```
2016-10-20 18:06:39.308606 [pkg_logger.go:134] I | keeper: id: postgres2
[...]

Issue #160, opened 2016-10-11
Move to a fully declarative model


sgotti:
To implement features like #144 #157 #159 some breaking changes in the current store model and also in keepers local db file handling are probably inevitable.

Use this breaking point to also move the store model to a full declarative model (similar to the k8s one).
There won't be a "real" api server like in k8s since it'll require implementing it inside the sentinels (to avoid adding additional processes) and making them a frontend between the clients and the store (this will require secure http calls, authentication, need to load balance calls to them, retry if a sentinel doesn't work and many other issues to handle). Instead the store writers (sentinels and stolonctl) will directly generate and validate the "api" data and write them to the store.

This will ease implementation of the above features, remove some overlapping and confusing logics and make future improvements easier to implement.

sgotti:
Implemented in #178

PR #172, opened 2016-10-21
FAQ section of README.md updated

afiskon:
Based on discussion: https://github.com/sorintlab/stolon/issues/168#issuecomment-255361446

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,46 @@ For our need to forcibly close connections to unelected masters and handle keepe
 
 We are open to alternative solutions (PRs are welcome) like using haproxy if they can met the above requirements. For example, an hypothetical haproxy based proxy needs a way to work with changing ip addresses, get the current cluster information and being able to forcibly close a connection when an haproxy backend is marked as failed (as a note, to achieve the latter, a possible solution that needs testing will be to use the [on-marked-down shutdown-sessions](https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.2-on-marked-down) haproxy server option).
 
+### Does proxy send read-only requests to standbys?
+
+Currently the proxy redirects all requests to the master. There is a [feature request](https://github.com/sorintlab/stolon/issues/132) for using the proxy also for standbys but it's low in the priority list. There is a workaround though.
+
+Application can learn cluster configuration from `stolon/cluster/mycluster/clusterdata` key. Consul allows to subscribe to updates of this key like this: 
+
+```
+http://localhost:8500/v1/kv/stolon/cluster/mycluster/clusterdata?wait=0s&index=14379
+```
+
+... where 14379 is ModifyIndex of a key reported by Consul.
+
+### How stolon decide which standby should be promoted to master?
+
+Currently it tries to find the best standby, the one with the xlog location nearest to the master latest knows xlog location. If a master is down there's no way to know its latest xlog position (stolon get and save it at some intervals) so there's no way to guarantee that the standby is not behind but just that the best standby of the ones available will be choosen. 
+
+### Does synchronous replication mean that I can't loose any data?
+
+Since version 9.6 PostgreSQL supports [synchronous replication to the quorum](https://www.postgresql.org/docs/9.6/static/runtime-config-replication.html#GUC-SYNCHRONOUS-STANDBY-NAMES). Unfortunately stolon doesn't support this feature yet and configures replication like this:
+
+```
+# on postgres1 server
+synchronous_standby_names = 'postgres2,postgres3'
+```
+
+According to PostgreSQL documentation:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,6 +75,46 @@ For our need to forcibly close connections to unelected masters and handle keepe
...
+According to PostgreSQL documentation:
+
+> Specifies a comma-separated list of standby names that can support synchronous replication, as described in Section 25.2.8. At any one time there will be at most one active synchronous standby; transactions waiting for commit will be allowed to proceed after this standby server confirms receipt of their data. The synchronous standby will be the first standby named in this list that is both currently connected and streaming data in real-time (as shown by a state of streaming in the pg\_stat\_replication view). Other standby servers appearing later in this list represent potential synchronous standbys.
+
+It means that in case of netsplit synchronous standby can be not among majority nodes. In this case some recent changes will be lost. Although it's not a major problem for most web projects, currently you shouldn't use stolon for storing data that under no circumstances can't be lost.
+
+### Does stolon use Consul as a DNS server as well?
+
+Consul (or etcd) is used only as a key-value storage.
+
```
sgotti: I don't completely get the meaning of this question. Do you mean registering a service in consul? If so which service (the proxies?)? I don't see why stolon should do this.

README.md
```diff
@@ -75,6 +75,46 @@ For our need to forcibly close connections to unelected masters and handle keepe
...
+# on postgres1 server
+synchronous_standby_names = 'postgres2,postgres3'
+```
+
+According to PostgreSQL documentation:
+
+> Specifies a comma-separated list of standby names that can support synchronous replication, as described in Section 25.2.8. At any one time there will be at most one active synchronous standby; transactions waiting for commit will be allowed to proceed after this standby server confirms receipt of their data. The synchronous standby will be the first standby named in this list that is both currently connected and streaming data in real-time (as shown by a state of streaming in the pg\_stat\_replication view). Other standby servers appearing later in this list represent potential synchronous standbys.
+
+It means that in case of netsplit synchronous standby can be not among majority nodes. In this case some recent changes will be lost. Although it's not a major problem for most web projects, currently you shouldn't use stolon for storing data that under no circumstances can't be lost.
+
```
sgotti: I won't add the concept of quorum here since it creates more confusion. Also the postgres doc doesn't talk about "quorum".

In addition the real problem here is not a netsplit (this is just one of the possible causes) but the fact that we let postgres choose the active synchronous standby, so the sentinel cannot know what was the active synchronous standby when the master was declared as dead. So the unique ways the sentinel has is to find the "best" standby based on the last know xlog position. But if both the master and the active synchronous standby goes down at the same time another standby will be choosed and it cannot be in full sync.

I opened #173 with a description and a solution to this. This will work also with postgresql <= 9.5 but with the limitation of setting only one sync standby. Thoughts?

README.md
```diff
@@ -75,6 +75,46 @@ For our need to forcibly close connections to unelected masters and handle keepe
 
 We are open to alternative solutions (PRs are welcome) like using haproxy if they can met the above requirements. For example, an hypothetical haproxy based proxy needs a way to work with changing ip addresses, get the current cluster information and being able to forcibly close a connection when an haproxy backend is marked as failed (as a note, to achieve the latter, a possible solution that needs testing will be to use the [on-marked-down shutdown-sessions](https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.2-on-marked-down) haproxy server option).
 
+### Does proxy send read-only requests to standbys?
+
+Currently the proxy redirects all requests to the master. There is a [feature request](https://github.com/sorintlab/stolon/issues/132) for using the proxy also for standbys but it's low in the priority list. There is a workaround though.
+
+Application can learn cluster configuration from `stolon/cluster/mycluster/clusterdata` key. Consul allows to subscribe to updates of this key like this:
```
sgotti: I'll add an example (also if this is going to change with #160) on what to do with that data (ie get the clusterview.keeperole infos)

The real problem with this that it's not assured that the standby are in sync, dead or else without more logic.

README.md
```diff
@@ -75,6 +75,46 @@ For our need to forcibly close connections to unelected masters and handle keepe
 
 We are open to alternative solutions (PRs are welcome) like using haproxy if they can met the above requirements. For example, an hypothetical haproxy based proxy needs a way to work with changing ip addresses, get the current cluster information and being able to forcibly close a connection when an haproxy backend is marked as failed (as a note, to achieve the latter, a possible solution that needs testing will be to use the [on-marked-down shutdown-sessions](https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.2-on-marked-down) haproxy server option).
 
+### Does proxy send read-only requests to standbys?
+
+Currently the proxy redirects all requests to the master. There is a [feature request](https://github.com/sorintlab/stolon/issues/132) for using the proxy also for standbys but it's low in the priority list. There is a workaround though.
+
+Application can learn cluster configuration from `stolon/cluster/mycluster/clusterdata` key. Consul allows to subscribe to updates of this key like this: 
+
```
sgotti: This is also available with etcd. Not sure if this should be detailed but could just say that one can use the watching features of the store.

sgotti:
After #219 I'll open a PR to update these FAQs and change them due to the new implemented features.

sgotti:
Reworked in #224.

PR #231, opened 2016-12-20
Stolon cluster with Docker Network.

ihcsim:
Not sure if this will be helpful, I added an example of running stolon cluster with Docker Network. More information can be found in the README.

I'm also currently testing it with Docker Compose; then later Swarm.

### DIFF
--- a/build
+++ b/build
@@ -94,10 +94,13 @@ rm -f ${BINDIR}/stolon-keeper
 cp ${GOPATH}/bin/keeper ${BINDIR}/stolon-keeper
 
 # Copy binaries to Dockerfile image directories
-DOCKERIMAGE=${BASEDIR}/examples/kubernetes/image/docker
-
-mkdir -p ${DOCKERIMAGE}/bin/
-rm -f ${DOCKERIMAGE}/bin/*
-for cmd in stolon-keeper stolon-sentinel stolon-proxy stolonctl; do
-	cp ${BINDIR}/${cmd} ${DOCKERIMAGE}/bin/
+declare -a DOCKERFILE_PATHS
+DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker ${BASEDIR}/examples/docker)
+for path in "${DOCKERFILE_PATHS[@]}"
+do
+  rm -rf $path/bin/
+  mkdir -p $path/bin/
+  for cmd in stolon-keeper stolon-sentinel stolon-proxy stolonctl; do
+          cp ${BINDIR}/${cmd} $path/bin/
+  done
 done
--- a/examples/docker/local-up.sh
+++ b/examples/docker/local-up.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+set -e
+
+usage() {
+  cat <<EOF
+The following is the list of required variables:
+  ETCD_VERSION         Etcd Docker image version.
+  ETCD_TOKEN           Initial cluster token for the etcd cluster during bootstrap.
+  IMAGE_TAG_SENTINEL   Stolon Sentinel Docker image tag.
+  IMAGE_TAG_KEEPER     Stolon Keeper Docker image tag.
+  IMAGE_TAG_PROXY      Stolon Proxy Docker image tag.
+  STOLON_PROXY_PORT    Port that Stolon Proxy listens on.
+EOF
+}
+
+if [ -z "$ETCD_VERSION" ]
+then
+  echo "Can't create etcd cluster. Missing required ETCD_VERSION."
+  usage
+  exit 1
+fi
+
+if [ -z "$ETCD_TOKEN" ]
+then
+  echo "Can't create etcd cluster. Missing required ETCD_TOKEN."
+  usage
+  exit 1
+fi
+
+if [ -z "$IMAGE_TAG_SENTINEL" ]
+then
+  echo "Can't create stolon cluster. Missing required IMAGE_TAG_SENTINEL."
+  usage
+  exit 1
+fi
+
+if [ -z "$IMAGE_TAG_KEEPER" ]
+then
+  echo "Can't create stolon cluster. Missing required IMAGE_TAG_KEEPER."
+  usage
+  exit 1
+fi
+
+if [ -z "$IMAGE_TAG_PROXY" ]
+then
+  echo "Can't create stolon cluster. Missing required IMAGE_TAG_PROXY."
+  usage
+  exit 1
+fi
+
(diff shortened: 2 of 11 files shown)


sgotti:
@ihcsim that's really great. I'll go through this PR and do a review.

> I'm also currently testing it with Docker Compose; then later Swarm.

Good. At a first sight looks like your example uses default docker functionalities but it could be easily expanded to use docker in [swarm mode](https://docs.docker.com/engine/swarm/) that will provide similar features to k8s for production workloads (like service discovery, container rescheduling etc...) (when using a volume driver that works correctly between multiple hosts).

ihcsim:
@sgotti I made some changes to the Dockerfiles in this PR to work with Docker Compose and Docker Service scaling:

1. Added an entrypoint script for Keeper to use the runtime `$HOSTNAME` environmental variable as the value of `--pg-listen-addr`. This ensures that all the Keeper replicas created by `docker-compose scale` and `docker service scale` are assigned different listening address. My previous approach of using the container name won't work with scaling because the replicas' addresses and names are unknown at build time.
1. Removed all unnecessary build arguments from the Dockerfiles.

sgotti:
@ihcsim Great! Do you think it's ready for a review?

ihcsim:
@sgotti Good to go. I'll rebase the branch with master once the review is completed.

ihcsim:
@sgotti I have feature branches for Docker Compose and Swarm mode example. Shall I merge them into one big PR? Or just wait for you to finish review this PR first? Might be easier to reason with smaller PRs.

build
```diff
@@ -94,10 +94,13 @@ rm -f ${BINDIR}/stolon-keeper
 cp ${GOPATH}/bin/keeper ${BINDIR}/stolon-keeper
 
 # Copy binaries to Dockerfile image directories
-DOCKERIMAGE=${BASEDIR}/examples/kubernetes/image/docker
-
-mkdir -p ${DOCKERIMAGE}/bin/
-rm -f ${DOCKERIMAGE}/bin/*
-for cmd in stolon-keeper stolon-sentinel stolon-proxy stolonctl; do
-	cp ${BINDIR}/${cmd} ${DOCKERIMAGE}/bin/
+declare -a $DOCKERFILE_PATHS
```
sgotti: This should probably be `declare -a DOCKERFILE_PATHS` or it will be expanded to `declare -a` and just show the current bash arrays.

examples/docker/local-up.sh
```diff
@@ -0,0 +1,94 @@
...
+    --initial-cluster-token $ETCD_TOKEN
+done
+
+docker run --name stolon-sentinel -d --net=$network -e STSENTINEL_STORE_ENDPOINTS=$etcd_endpoints $IMAGE_TAG_SENTINEL
+
+declare -a keepers
+keepers=(stolon-keeper-00 stolon-keeper-01 stolon-keeper-02)
+for keeper in "${keepers[@]}"
+do
+  docker run --name $keeper -d --net=$network -e STKEEPER_STORE_ENDPOINTS=$etcd_endpoints $IMAGE_TAG_KEEPER
```
sgotti: Setting also `STKEEPER_UID` to `$keeper` instead of using a keeper generated random uid will make the logs and stolonctl output cleaner (easier to get the current master keeper)

ihcsim:
@sgotti Thanks for the feedback.

> It would be good to just use one image for all the components like done with the k8s example (perhaps the same image).

What are the benefits of using one image? The different Dockerfiles show that these images are quite different. For example,

1. Sentinel needs an init cluster spec file.
1. Keepers, as you pointed out, have data directory, secrets and volumes.
1. Their environmental variables are different.

A combined image is also bigger because it has binaries that it doesn't use.

examples/docker/local-up.sh
ihcsim: Though I can use specific UID for keeper for this example, I won't be able to do that for Compose and Swarm if I wanted to scale the keepers. Otherwise, the replicas will end up with the same `STKEEPER_UID`.

build
ihcsim: Good catch! Fixed.

examples/docker/local-up.sh
sgotti: ok. I'm not sure how docker swarm replicas work and if they provide persistent volumes and persistent names always tied to the same container. In k8s we achieve this using a statefulset (petset) or creating N replication controllers with replicas set to 1.

sgotti:
> What are the benefits of using one image? The different Dockerfiles show that these images are quite different. For example,
> 
> Sentinel needs an init cluster spec file.
> Keepers, as you pointed out, have data directory, secrets and volumes.
> Their environmental variables are different.
> A combined image is also bigger because it has binaries that it doesn't use.

With k8s we achieve this since we can pass secrets, persistent volumes, entrypoints etc... in the definitions, so we have one single image that just provides the binaries.

But I'm not sure if doing the same with docker/docker swarm/docker compose is straightforward as on kubernetes and if it's a best practice or just using multiple images is the preferred way.

ihcsim:
> But I'm not sure if doing the same with docker/docker swarm/docker compose is straightforward as on kubernetes and if it's a best practice or just using multiple images is the preferred way.

The images built by the Dockerfile in the Kubernetes example aren't very useful on their own. Say if you post them to DockerHub, and a user do a `docker pull` on those images, there are no easy way to tell what arguments are needed to get `docker run` to work. If your preference is to re-use the Kubernetes example Dockerfile, then feel free to just change the [local-up.sh](https://github.com/ihcsim/stolon/blob/935b05d296abcd5fa302209870cbfbd8f98650c4/examples/docker/local-up.sh) script accordingly.

sgotti:
> The images built by the Dockerfile in the Kubernetes example aren't very useful on their own.

This is how all the k8s examples works, the image provides the needed binaries. Instead all the configurations, entrypoints are defined in the provided descriptors. A lot of people argue that mixing both the "how" an image is built and "what" it needs to run in the same Dockerfile is not a good idea.

Also publishing an image containing secrects is not feasible so you'll always have to take it and build a new image on top of it with the additional files and run it providing external volumes and additional configuration.

But I don't want to block you since I think your examples are really good and appreciated. I'll go ahead merging this and also the future docker swarm and compose examples and in future we'll see how things will change and adjust the examples.

ihcsim:
Sounds good. Yeah, feel free to adjust and change the examples as you see fit. Once you merged this PR, I will rebase the other compose/swarm branch with your master HEAD, before opening a new PR for that.

sgotti:
@ihcsim Sorry for the delay. Merged. Thanks a lot!

Issue #232, opened 2016-12-23
missing flock on unsupported OS (solaris)


siepkes:
I tried to compile stolon for use on SmartOS (Illumos / Solaris) an ran into an issue. The issue is that CoreOS rkt (on which Stolon is dependent) uses the `flock` system call. Apparently Illumos now supports flock (see: [Feature #3252](https://www.illumos.org/issues/3252)) however to Go Solaris library does not yet support it. Long story short; Solving this upstream in Go is obviously the right way (which is apparently also being worked on) but thats going to take a looooooong time before being complete. 

Now I noticed that the only reason why stolon is (currently) dependent on rkt is because of a lock implementation which is used in `keeper.go` line 1406 and nothing else. Would you be open to ditching the rkt dependency and using a lock implementation which also works on Illumos / Solaris? rkt also seems like quite a big "library" to use for just a small lock implementation.

For Google-ing reference, this is the error Go dies with when trying to compile Stolon (with rkt) for Solaris:
```
# github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:49: undefined: syscall.Flock
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:49: undefined: syscall.LOCK_EX
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:49: undefined: syscall.LOCK_NB
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:75: undefined: syscall.Flock
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:75: undefined: syscall.LOCK_EX
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:96: undefined: syscall.Flock
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:96: undefined: syscall.LOCK_SH
gopath/src/github.com/sorintlab/stolon/vendor/github.com/coreos/rkt/pkg/lock/file.go:96: undefined: syscall.LOCK_NB
[...]

PR #236, opened 2017-01-17
Use POSIX compatible locking with fcntl() instead of flock()

siepkes:
As discussed in issue #232 I tried using go4/lock (even created a pull request for native Solaris locking support: camlistore/go4#26) but bumped into an issue. By default go4/lock assumes you are locking a file. Trying to lock a directory doesn't work. Modifying go4/lock to accomodate locking directories would be a bit more work (also ensuring the fallback locking works on directories). Supporting the fallback method would also introduce another complication; We would need to cleanly release the lock at shutdown. This isn't needed for fcntl or flock since the lock is gone after the processes exists. However when using the file based fallback lock the fallback file would need to be cleaned up. Because os.Exit is called from various places in keeper.go releasing the lock cleanly would require some refactoring.

So I thought we could just use locking based on `fcntl` instead. `fcntl` is part of POSIX so it is widely supported (Linux, all of the BSD's, OS X, Solaris, Cygwin, etc.). Using fnctl() over the existing implementation which uses flock is an improvement IMHO because `fnctl()` is part of POSIX and `flock()` is not. From what I could find online is that flock is basically a left over from the BSD-era.


siepkes:
I noticed the CI bailed out on `undefined: io.SeekStart`. I used Go 1.7.4 but the CI seems to be 1.5.4. Is there a specific Go version I should target?

sgotti:
@siepkes 
Looks like you forgot to gofmt the code.

> I noticed the CI bailed out on undefined: io.SeekStart. I used Go 1.7.4 but the CI seems to be 1.5.4. Is there a specific Go version I should target?

Probably we should move the only the latest 2 go versions. For the moment you can just use `0`  instead of `io.SeekStart`

sgotti:
@siepkes sorry for being late but I was waiting for a go for the review. Looks like TestExclusiveLock is failing.

I think you're calling the fcntl syscall with the wrong parameters. You could take a look at how go4/lock calls it: https://github.com/camlistore/go4/blob/master/lock/lock_unix.go

sgotti:
@siepkes I forgot that you cannot take a posix lock on a directory but only on a file. So you should
`os.Create` (that will create a file if it doens't exists or open and truncate if it exists) instead of the current `os.Open` a file named "lock" inside the datadir.

siepkes:
@sgotti Sorry I've been off the grid for a while. I'm still interested in this and i'll take a look at this the week after next week (because I've got a small vacation planned to Rome wooohoooo! ;-)

siepkes:
@sgotti So instead of trying to lock a directory I now lock a file inside the data directory. I also squashed all commits in the pull requests to make the whole thing more readable. However the integration tests still fail. I'm having a hard time determining which test actually failed and why my changes would actually cause them to fail.

sgotti:
@siepkes the failing test is `TestExclusiveLock`, since you changed the error output on failed lock acquisition it's expecting a different one: https://github.com/sorintlab/stolon/blob/master/tests/integration/init_test.go#L505

cmd/keeper/keeper.go
```diff
@@ -1463,11 +1462,29 @@ func keeper(cmd *cobra.Command, args []string) {
...
-	_, err = lock.TryExclusiveLock(cfg.dataDir, lock.Dir)
+	// Open (and create if needed) the lock file.
+	// There is no need to clean up this file since we don't use the file as an actual lock. We get a lock
+	// on the file. So the lock get released when our process stops (or dies).
+	lockFileName := filepath.Join(cfg.dataDir, "lock")
+	lockFile, err := os.OpenFile(lockFileName, os.O_RDWR|os.O_CREATE, 0644)
 	if err != nil {
-		die("cannot take exclusive lock on data dir %q: %v", cfg.dataDir, err)
+		fmt.Printf("cannot open lock file %q: %v\n", lockFileName, err)
+		os.Exit(1)
```
sgotti: let's use `die`

glide.yaml
```diff
@@ -1,9 +1,5 @@
 package: github.com/sorintlab/stolon
 import:
-- package: github.com/coreos/rkt
-  version: v1.16.0
-  subpackages:
-  - pkg/lock
```
sgotti: you should also remove the directory inside vendor: `vendor/github.com/coreos/rkt`

siepkes:
@sgotti Green across the board! I reverted to the old error message (that should prevent anyone else's workflow from breaking who parses that text) and squashed all commits to a single commit. I also used `die` instead of print and exit.

sgotti:
@siepkes Great! I'll merge in a few minutes. Thanks a lot!

sgotti:
@siepkes Just noticed that we should also rebuild the glide.lock file, but since we are vendoring the files it's not a big issue and I'll do this in another PR.

siepkes:
@sgotti Thanks!

PR #248, opened 2017-03-13
Write files atomically, fixes #247

### LINKED ISSUE #247: dbstate file can end up empty
I encountered a situation where the file `/path/to/dbstate` was empty (0 bytes length) and `stolon-keeper` would emit:

```
cannot create keeper: failed to load db local state file: unexpected end of JSON input
```

Is it possible that this happens if the keeper dies at the right moment? Maybe it should use an atomic move (write to a temp file first, then `mv` to the target location) to avoid this?

nh2:


sgotti:
Thanks for the PR, I'll review it tomorrow!

common/common.go
```diff
@@ -66,3 +69,33 @@ type Parameters map[string]string
 func (s Parameters) Equals(is Parameters) bool {
 	return reflect.DeepEqual(s, is)
 }
+
+// Write file to temp and atomically move when everything else succeeds.
+// This function is taken from
+//   https://github.com/youtube/vitess/blob/master/go/ioutil2/ioutil.go
+// Copyright 2012, Google Inc. BSD-license
```
sgotti: I'm not expert but I hope this is enough to respect google bsd like license (or should we add a special case in our Apache 2 license file?)

nh2: It would be slightly better to drop a copy of the BSD license somewhere into the repo if you're OK with that (how about a `licenses` directory) and reference this from here, as conceptually you are supposed to include the full text of the license somewhere (because it carries the disclaimer).

sgotti: @nh2 I like your proposal of adding a licenses directories containing the BSD one and point the comment to it.

sgotti:
superseded by #265

PR #249, opened 2017-03-16
pipe postgres output to std[err|out] fixes #244

### LINKED ISSUE #244: Postgres stderr logs as part of stolon-keeper
I am running `stolon-keeper` as part of a docker image, and I cannot find a way to see logs (via the command `docker logs`) of both `stolon-keeper` and the underlying `postgres` process. Only the logs related to the operation of `stolon-keeper` are shown.
(I can of course `ssh` into the container and read the logs from there just fine, and I made sure that the postgres parameter`log_destination` is set to `stderr`)

Is there anything I should pay attention to?

sajal:
I am evaluating stolon on Kubernetes (largely based on examples provided here). Having access to postgres logs is nice... If this is not the desired default behaviour, it could be hidden behind a flag.


pkg/postgresql/postgresql.go
```diff
@@ -193,12 +193,9 @@ func (p *Manager) start(args ...string) error {
...
 	log.Debug("execing cmd", zap.Object("cmd", cmd))
-	// TODO(sgotti) attaching a pipe to sdtout/stderr makes the postgres
-	// process executed by pg_ctl inheriting it's file descriptors. So
-	// cmd.Wait() will block and waiting on them to be closed (will happend
-	// only when postgres is stopped). So this functions will never return.
-	// To avoid this no output is captured. If needed there's the need to
-	// find a way to get the output whitout blocking.
+	//Pipe command's std[err|out] to parent.
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
```
sgotti: @sajal I don't completely remember the reason why I wrote that comment but looks like piping the cmd output to the stdout/stderr is enough.

sgotti:
@sajal Thanks for the PR, I was thinking of doing some similar.

The unique "issue" is that it's difficult to distinguish postgres logs from the keeper logs. A solution I was thinking about (but perhaps it could be done in another PR) will be to use two goroutines to continuously read postgres stdout and stderr and pass their output to the logging functions so we can specifiy it's the postgres out, do structured logging etc...

sajal:
I have seen some log package (not sure standard library or 3rd party) provide an `io.Writer` interface. Will look more into it.

sajal:
I am prototyping some change to feed the std[err|out] into zap with a prefix, but wont be able to properly test till tomorrow.

pkg/postgresql/postgresql.go
```diff
@@ -44,6 +44,23 @@ func SetLogger(l zap.Logger) {
...
+}
+
+func NewZapLogWriter(tag string, logger *zap.Logger) *LogWriter {
+	lw := &LogWriter{}
+	lw.tag = tag
+	return lw
+}
+
+func (lw ZapLogWriter) Write(p []byte) (n int, err error) {
+	lw.logger.Info(lw.tag + ": " + string(p))
```
sajal: @sgotti : I am not sure of this part, not familiar with zap

sajal:
Apologies for the commit spam.

sgotti:
@sajal don't worry, push as nuch commits as you want. :smile:

The issue I see (and have to think about it a bit more) is that setting the postgres stdout/err to the keepers ones will work and cmd.Wait will exit when pg_ctl exits, instead passing an io.Writer will make cmd.Start setup two goroutines copying the data from the child process stdout/err to the io.Writer and will probably make cmd.Wait block (or two goroutines will be leaked).

Perhaps we can just start with your first patch and improve it if possible in a future PR.

sajal:
@sgotti There are 3 approaches as I see it.

1. Just pass the cmd's output to os.Std[err|out]. Looses contextual information, but assume any log without context is postgres. - The original PR.
2. Set io.Writer to cmd's output and send it to zap. 
3. Use cmd.Std[out|err)Pipe() to get a reader and launch 2 goroutinues to read it and write to logs.

I don't see option 2 leaking any goroutinues, but it might make cmd.Wait/Run block, if the logger implementation has any blocking things

Option 3 seems to be what you want, but the [docs](https://golang.org/pkg/os/exec/#Cmd.StderrPipe) say to consume all the logs before calling `Wait`. That does not sound practical. 

Can we go ahead with the original PR (Option 1) for now and treat "distinguish postgres logs" as enhancement? For now I just wanna be able to see what postgres is doing.

sgotti:
@sajal Sure, let's go ahead with your initial patch. I'll reopen this PR, you can just force push your first commit.

sajal:
@sgotti Done

sgotti:
@sajal Merged. Thanks a lot! I'll open a follow up issue related to piping the output to the keeper logger.

PR #257, opened 2017-04-04
Handle SIGTERM as well as SIGINT for shutdown.

Dirbaio:
The current keeper code is handling only SIGINT for shutdown. If the keeper receives a SIGTERM it shuts down uncleanly. If running inside a Docker container, this means the Postgres process is killed uncleanly as well, which is bad.

This is important because Docker's `docker stop` and Kubernetes' `kubectl delete pod` stop the container by sending SIGTERM. 

This PR: 
- Adds SIGTERM to the handler
- Removes SIGKILL, since it can't be catched.
- Changes SIGINT to use syscall as well, for consistency.

### DIFF
--- a/cmd/keeper/keeper.go
+++ b/cmd/keeper/keeper.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/sorintlab/stolon/common"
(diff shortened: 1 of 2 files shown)


sgotti:
@Dirbaio good catch, the real intentions was to trap SIGINT and SIGTERM but my intent to rely only on the os package made it become SIGKILL. Thanks a lot!

PR #265, opened 2017-04-23
keeper: write files atomically

### LINKED ISSUE #247: dbstate file can end up empty

sgotti:
fixes #247

### DIFF
--- a/cmd/keeper/keeper.go
+++ b/cmd/keeper/keeper.go
@@ -1359,7 +1359,7 @@ func (p *PostgresKeeper) saveKeeperLocalState() error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


sgotti:
@nh2 changed the comment to adhere to golang function comment style and added the license as discussed in #248

nh2:
Thanks!

PR #299, opened 2017-06-22
Fix typo in initialization log line

theckman:
In the `initCluster()` function there was a log line to `stdout` where we
warn the user that the database data will be overwritten. The word overwritten
was mis-typed as overwrited. This change fixes that.

Signed-off-by: Tim Heckman <t@heckman.io>


sgotti:
@theckman Thanks a lot! Merging.

PR #301, opened 2017-06-28
stolonctl: add remove keeper command

sergik:
add remove keeper cmd to stolonctl. 
`stolonctl remove <keeperId>`
Useful for removing dead keepers from stolon config. Removes keeper info and associated db info from stolon config.


sergik:
I there some infrastructure problems or smth should be fixed from my side?
curl: (35) Unknown SSL protocol error in connection to releases.hashicorp.com:443

cmd/stolonctl/removeKeeper.go
```diff
@@ -0,0 +1,84 @@
...
+
+	keeperDb := getDbForKeeper(newCd.DBs, keeperID)
+
+	if keeperDb == nil {
+		die("Can not find db for specified keeper id")
+	}
+
+	delete(newCd.DBs, keeperDb.UID)
+	delete(newCd.Keepers, keeperID)
+
```
sgotti: I think that some restrictions should be added. Currently I can think of this important one:

* The current master keeper should never be removed or the whole cluster will break because in normal state there must always be a master keeper.

Some notes:

* If a keeper is marked as alive removing it will make the associated db be removed but if it's still alive it will reregister

sgotti:
> I there some infrastructure problems or smth should be fixed from my side?
> curl: (35) Unknown SSL protocol error in connection to releases.hashicorp.com:443

probably a transient network error.

sergik:
Added check for current master

sgotti:
@sergik thanks for the update, I'll be more comfortable merging this if you can add some integration tests. Some cases I'd like to cover could be:

* Master removal (should fail)
* Synchronous standby removal (to be sure this doesn't break current synchronous replication logic).

If you have some problems let me know or if you don't have time I'll write them and create a new PR with your commit and the added tests.

sgotti:
@sergik gentle ping. Can I do something to move on this? Thanks!

sgotti:
closing in favor of #383

PR #304, opened 2017-07-04
tests: increase some timeouts

sgotti:
sometime some tests timeouts on semaphore (since we are using a lot of
parallel jobs and some processes can become starved on I/O).


sgotti:
closing since the branch name is not right.

PR #315, opened 2017-07-14
scripts: Use `/usr/bin/env bash` instead of `/bin/bash`.

nh2:
Same approach as is common for python.

Fixes builds on NixOS and other environments that guarantee
reproducible builds.


sgotti:
@nh2 LGTM. Thanks!

PR #326, opened 2017-07-27
Replaced sprintf with net.JoinHostPort to fix bug with ipv6

emdem:


sgotti:
@emdem Thanks for the PR.

I honestly never tried stolon with ipv6 addresses so if only this change is needed to make it work I'll merge it.

emdem:
@sgotti No problem, thanks for open-sourcing! It may not be the only one, but I'll fix them as I find them. I think this covers the proxy at least.

emdem:
@sgotti I've confirmed that it fixes the issue I was having with the proxy. The following all work:
ipv4 proxy -> ipv6 keeper
ipv6 proxy -> ipv6 keeper
ipv6 proxy -> ipv4 keeper

sgotti:
@emdem Great! We should probably add more tests for ipv6 in future. Merging. Thanks a lot!

PR #350, opened 2017-09-04
postgres: also fetch wals during pg_basebackup

sgotti:
Also fetch wals when calling pg_basebackup. We could also stream them but this
will require an additional connection and wal sender so for now, since we are
using replication slots, just rely on fetching at the end of the basebackup.

This will fix pg_basebackup waiting indefinetly if archive mode is enabled but
the archive command on the primary isn't working correctly.


vitalif:
Hi, maybe you could implement basebackup WAL streaming mode?
Because currently it's possible that slaves can't finish basebackup when master is loaded with write requests.
Or maybe it is already implemented? :)

Issue #290, opened 2017-06-04
Tablespaces support


sgotti:
Now stolon doesn't works with tablespaces since it can't know where the tablespaces files will be placed and cannot remove them during resync.

Some possibilities:

* Add a way to define a list of tablespaces dirs in the cluster data
* A preresync script to be executed before pg_basebackup where user can add some cleanup commands.

smkingsoft:
Maybe preresync script can give user more customization.

smkingsoft:
@sgotti 

Can use simple string array define User TableSpaces dir root
eg:
stolonctl init '{ "initMode": "new", "pgParameters": {"log_min_duration_statement": "1s" }, "pgUserTbSps": ['/UserData1','UserData2']}'

The User table space path is /UserData1/UserTb1, /UserData1/UserTb2,  /UserData2/UserTb3, /UserData2/UserTb4

When stolon is initialized, the directory '/UserData1' ,'/UserData2'  is created

sgotti:
@smkingsoft Defining a static list of directories is the faster way. The keeper will delete and recreate them when doing a full resync. The admin must update this list before adding new tablespaces in additional directories.

An improvement could be to automatically retrieve them (the keeper will get them and publish them to the cluster data like already done with other information like pg parameters) so the admin shouldn't need to set them manually.

But I think there are some unfixable corner cases caused by the pg tablespaces nature: they can be defined everywhere outside the pg data dir and so stolon won't be able to fully control them.

For example imagine a standby that has a tablespace directory that doesn't exists, has wrong permissions or is filled with other data while on the master it's all ok, if we create a new tablespace on the master it will succeed while the slave will probably (haven't tested it) break since it cannot create the data files.

smkingsoft:
@sgotti 
In an online system, it is possible to create a tablespace directory outside the pg data dir. 
the DBA will set the directory and permissions in advance,   the directory will fill only pg data it is his work.
My idea is to set the root directory of the tablespace directory, the tabelspace dir is subdir in the root,  as long as the root directory does not change,  the administrator does not care the tablespace add or del.

sgotti:
@smkingsoft I think your proposal will add another concept not existing in postgres: tablespaces root,  the unique usefulness of it is to avoid specifing all the tablespaces dirs but it'll also put some conditions on how users should define their tablespaces structures.

[...]

PR #353, opened 2017-09-15
add tablespaces support

smkingsoft:
hi @sgotti 
when resync befor SyncFromFollowed function

i use sql "select spcname as name, pg_catalog.pg_tablespace_location(oid) as location from pg_catalog.pg_tablespace order by 1"  query from master db

then remove all user tablespaces dir and recreate these


smkingsoft:
i first use pullrequst, may be has some mistake.

sgotti:
@smkingsoft thanks for the PR! I'll review it in the next days.

sgotti:
@smkingsoft Can you please gofmt the files (that's the cause of the failed tests). There's also the need of some integration tests.

But, like explained in https://github.com/sorintlab/stolon/issues/290#issuecomment-323286495 tablespaces can cause problems since we cannot control tablespaces paths, so I'll also document and warn about the possible problems when using tablespaces inside stolon.

smkingsoft:
OK, i will try gofmt. But i first use go, so I do not know how to write tests.
could you give me some suggest? @sgotti 

Yes, use the table space to be careful!

sgotti:
@smkingsoft see https://github.com/sorintlab/stolon/tree/master/tests/integration. If you think that these are too complex I can see if I can find the time to write some to test your PR. 

> Yes, use the table space to be careful!

We should add some docs stating the possible problems when using tablespaces.

smkingsoft:
@sgotti hi, How to Update Pull？
request a New pull?  or Update this pull?

sgotti:
@smkingsoft  You can just rebase and squash the commits and force push them to your branch.

vsbace:
Hi, all,
On our setup we have many non default tablespaces, how much I understood from different threads on this issue, it was never solved what to do to solve it, use some init PG params or something else?

PR #368, opened 2017-10-12
proxy: vendor updated pollon

### LINKED ISSUE #365: Long timeout for client when master goes down
If a client try to open a connection just after the master goes down, the client may experience a timeout of 2 minutes.

It should be expected that proxy abort the connection as soon as a new master is elected.

Step to reproduce this issue on minikube: https://gist.github.com/PierreF/cbde2755ce460b58b524707c554b61d1

I think the cause is that [net.Dial](https://github.com/sorintlab/stolon/blob/149980c23f193989f7b08eeb1a1f35bda420042d/vendor/github.com/sorintlab/pollon/pollon.go#L71) hang for up to 2 minutes.

net.Dial will hang because:
* The master was up, thus the ARP address is present in arp cache
* When the master goes down (here it's kubectl delete po/...), the IP goes unassigned and the ARP address also. But stay present in arp cache
* When a client try to open a connection, the proxy will net.Dial() the IP of the now down master. Since arp cache is filled, this result in a TCP/IP packet being send to that IP. Nothing will response to this packet, because the address is unassigned.
* TCP will retry sending this TCP/IP packet, while the arp cache is filled, no error will occur.
* As soon as the arp cache is purged and TCP make another retry, no an "host unreachable" error will be raised and net.Dial will fail.

By making net.Dial cancellable (https://github.com/sorintlab/pollon/pull/8), the client fail as soon as the new master is elected as expected.

sgotti:
Fixes #365

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 4fd509083e97727c4b09d5e4082f82bdb8e7dcdc607d7c49e019ee967c5802b4
-updated: 2017-09-21T17:59:18.467613065+02:00
+hash: 65066aa06af019a116ae79b13a0758b881c0158c8a4c341d7ffc021de3a4ff7a
+updated: 2017-10-12T18:04:47.972619241+02:00
 imports:
 - name: github.com/coreos/etcd
   version: bc9ddf260115d2680191c46977ae72b837785472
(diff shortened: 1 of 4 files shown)


sgotti:
@PierreF Do you prefer if I put you as the author if this commit?

PierreF:
Do the easiest for you, it doesn't matter to me.

PR #371, opened 2017-10-16
Add example for major postgresql upgrade on k8s

PierreF:
Provide an example of upgrade from PostgreSQL 9.6 to PostgreSQL 10.0 when running on Kubernetes.


sgotti:
@PierreF Thanks for the PR and sorry for being late. I haven't tried it but it looks good to me.

PR #379, opened 2017-11-14
Add waitFullRecovery option to PITR

PierreF:
When restoring with PITR (at least with WAL-E), stolon-keeper do not wait until full recovery before marking itself as initialized.
This allow for read-only connection on a database that is still recovering and thus allow client to see outdated content.

Step to reproduce:
* Configure a cluster with WAL-E backup
* Create a custom table, insert data
* Generate WAL, using pgbench for example
* Update data from custom table
* Trigger a PITR recovery. To help seeing the issue, adding a sleep in the restoreCommand may help
* During the recovery, once Postgres reach consistent state, stolon will accept connection. Querying the custom table will allow to see initial inserted data.

I've made this an option because I'm not sure if it was intended or not to allow read-only connection during the recovery.

### DIFF
--- a/cmd/keeper/keeper.go
+++ b/cmd/keeper/keeper.go
@@ -1073,7 +1073,14 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 				log.Errorw("failed to start instance", zap.Error(err))
 				return
 			}
-			// wait for the db having replyed all the wals
+
+			if standbySettings == nil {
+				// wait for the db having replyed all the wals
+				if err = pgm.WaitRecoveryDone(cd.Cluster.DefSpec().SyncTimeout.Duration); err != nil {
+					log.Errorw("recovery not finished", zap.Error(err))
+					return
+				}
+			}
 			if err = pgm.WaitReady(cd.Cluster.DefSpec().SyncTimeout.Duration); err != nil {
 				log.Errorw("instance not ready", zap.Error(err))
 				return
--- a/pkg/postgresql/postgresql.go
+++ b/pkg/postgresql/postgresql.go
@@ -341,6 +341,21 @@ func (p *Manager) WaitReady(timeout time.Duration) error {
 	return fmt.Errorf("timeout waiting for db ready")
 }
 
+func (p *Manager) WaitRecoveryDone(timeout time.Duration) error {
+	start := time.Now()
+	for time.Now().Add(-timeout).Before(start) {
+		_, err := os.Stat(filepath.Join(p.dataDir, "recovery.done"))
+		if err != nil && !os.IsNotExist(err) {
+			return err
+		}
+		if !os.IsNotExist(err) {
+			return nil
+		}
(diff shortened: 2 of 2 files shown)


PierreF:
I'm not sure why CI is failing, I don't understand what the error message is.
Could this be a temporary error ?

The failing test seems to be TestFailover, but my PR only change PITR process, not failover.

cmd/keeper/keeper.go
```diff
@@ -1073,8 +1073,16 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 				log.Errorw("failed to start instance", zap.Error(err))
 				return
 			}
-			// wait for the db having replyed all the wals
-			if err = pgm.WaitReady(cd.Cluster.DefSpec().SyncTimeout.Duration); err != nil {
+
+			deadline := time.Now().Add(cd.Cluster.DefSpec().SyncTimeout.Duration)
```
sgotti: deadline should be a time.Duration. You don't have to make it a time.Time adding it to time.Now() and then subtracting Time.Now in the below calls.

sgotti:
@PierreF Thanks for the PR! Comments in line.

I think this was an oversight, I haven't considered the case that postgres could be open before completing recovery. I think we could just apply the wait for recovery.done by default without any option.

> I'm not sure why CI is failing, I don't understand what the error message is.

Sometime due to parallel tests/high concurrency to speed up tests there's are some flakes. Will relaunch them if this happens.

PierreF:
Updated the PR to no longer be an option but enable by default.
Removed the deadline variable. It was a time.Time to use the same duration for both wait for recovery.done and wait for instance Ready. But if we wait for recovery.done (e.g. non-standby cluster), the instance should be ready when recovery.done is created and the timeout should not matter.

cmd/keeper/keeper.go
```diff
@@ -1073,7 +1073,14 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 				log.Errorw("failed to start instance", zap.Error(err))
 				return
 			}
-			// wait for the db having replyed all the wals
+
+			if standbySettings == nil {
```
sgotti: This check means that we'll waitRecoveryDone only when not syncing from an external instance (standby cluster). Is this really needed or it should be always done?

sgotti:
@PierreF Sorry but I lost your last update, just added another comment.

cmd/keeper/keeper.go
PierreF: I've never setup standby cluster, but I've guessed that recovery.done would never be created by PostgreSQL (since recovery should run forever until promoted to master).
If that guess is true, we should not wait here (or not with a timeout).

Also it seems desirable to be able to read from a standby cluster (e.g. for analytics) even if data is not the latest one.

sgotti: @PierreF You're right, I forgot that this is the pitr recovery case.

Issue #362, opened 2017-10-09
keeper: support different auth methods


sgotti:
Now stolon by default requires md5 password authentication for superuser and replication connections.

We could improve it to accept different authentication methods starting from the simpler ones like trust, md5, password, pam.

A solution will be to add some keeper options `--pg-su-auth-method` and `--pg-repl-auth-method`. These options will basically affect the pg_hba.conf generated replication entries and the user creation when initializing a new pg db cluster.

emdem:
@sgotti Which auth methods would you like to support initially? I was thinking trust, password, md5...

If that seems reasonable, I will test my implementation and make a PR. If not, please let me know which other auth methods you think would be reasonable for a first implementation.

Issue #42, opened 2015-10-18
Reject PostgreSQL instances that are below 9.4


thedrow:
stolon should not allow to connect to database that do not follow the minimum PostgreSQL requirement.

PR #380, opened 2017-11-14
Added auth method for su/replication user

emdem:
For issue #362

### DIFF
--- a/cmd/keeper/keeper.go
+++ b/cmd/keeper/keeper.go
@@ -81,9 +81,11 @@ type config struct {
 	pgListenAddress         string
 	pgPort                  string
 	pgBinPath               string
+	pgReplAuthMethod        string
 	pgReplUsername          string
 	pgReplPassword          string
 	pgReplPasswordFile      string
+	pgSUAuthMethod          string
 	pgSUUsername            string
 	pgSUPassword            string
 	pgSUPasswordFile        string
@@ -112,9 +114,11 @@ func init() {
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgListenAddress, "pg-listen-address", "localhost", "postgresql instance listening address")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgPort, "pg-port", "5432", "postgresql instance listening port")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgBinPath, "pg-bin-path", "", "absolute path to postgresql binaries. If empty they will be searched in the current PATH")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.pgReplAuthMethod, "pg-repl-auth-method", "md5", "postgres replication user auth method. Default is md5.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgReplUsername, "pg-repl-username", "", "postgres replication user name. Required. It'll be created on db initialization. Must be the same for all keepers.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgReplPassword, "pg-repl-password", "", "postgres replication user password. Only one of --pg-repl-password or --pg-repl-passwordfile must be provided. Must be the same for all keepers.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgReplPasswordFile, "pg-repl-passwordfile", "", "postgres replication user password file. Only one of --pg-repl-password or --pg-repl-passwordfile must be provided. Must be the same for all keepers.")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUAuthMethod, "pg-su-auth-method", "md5", "postgres superuser auth method. Default is md5.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUUsername, "pg-su-username", user, "postgres superuser user name. Used for keeper managed instance access and pg_rewind based synchronization. It'll be created on db initialization. Defaults to the name of the effective user running stolon-keeper. Must be the same for all keepers.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPassword, "pg-su-password", "", "postgres superuser password. Only one of --pg-su-password or --pg-su-passwordfile must be provided. Must be the same for all keepers.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPasswordFile, "pg-su-passwordfile", "", "postgres superuser password file. Only one of --pg-su-password or --pg-su-passwordfile must be provided. Must be the same for all keepers)")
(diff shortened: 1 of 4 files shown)


cmd/keeper/keeper.go
```diff
@@ -112,9 +114,11 @@ func init() {
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgListenAddress, "pg-listen-address", "localhost", "postgresql instance listening address")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgPort, "pg-port", "5432", "postgresql instance listening port")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgBinPath, "pg-bin-path", "", "absolute path to postgresql binaries. If empty they will be searched in the current PATH")
+	cmdKeeper.PersistentFlags().StringVar(&cfg.pgReplAuthMethod, "pg-repl-auth-method", "md5", "postgres replication user auth method. Required.")
```
sgotti: I'll remove "Required" since the default is set to md5 and will break current behavior.

cmd/keeper/keeper.go
```diff
@@ -1610,6 +1619,12 @@ func keeper(cmd *cobra.Command, args []string) {
 		die("cannot create data dir: %v", err)
 	}
 
+	if cfg.pgReplAuthMethod == "" {
+		die("--pg-repl-auth-method is required")
```
sgotti: Don't make it required. If not specified will used the default md5 as of now.

cmd/keeper/keeper.go
```diff
@@ -1579,6 +1587,7 @@ func main() {
 
 func keeper(cmd *cobra.Command, args []string) {
 	var err error
+	validAuthMethods := map[string]bool{"md5": true, "password": true, "trust": true}
```
sgotti: not a big deal but when a map doesn't require a values I prefer to use `struct{}`: `map[string]struct{}`

cmd/keeper/keeper.go
```diff
@@ -1620,6 +1635,11 @@ func keeper(cmd *cobra.Command, args []string) {
 	if cfg.pgReplPassword != "" && cfg.pgReplPasswordFile != "" {
 		die("only one of --pg-repl-password or --pg-repl-passwordfile must be provided")
```
sgotti: When using the trust auth method the user doesn't have to provide a password.

pkg/postgresql/postgresql.go
```diff
@@ -76,16 +78,18 @@ type InitConfig struct {
 	DataChecksums bool
 }
 
-func NewManager(pgBinPath string, dataDir string, localConnParams, replConnParams ConnParams, suUsername, suPassword, replUsername, replPassword string, requestTimeout time.Duration) *Manager {
+func NewManager(pgBinPath string, dataDir string, localConnParams, replConnParams ConnParams, suAuthMethod, suUsername, suPassword, replAuthMethod, replUsername, replPassword string, requestTimeout time.Duration) *Manager {
```
sgotti: In https://github.com/emdem/stolon/blob/8635f157ab57190d8607c179b8b66b47af7f7d6b/pkg/postgresql/postgresql.go#L363 we also populate the su and repl user password. I'll change them to populate it only with md5 auth since with trust I'll ignore if the user have provided a password.

cmd/keeper/keeper.go
sgotti: I'm not sure if we should also add the password method since postgres doc suggests to use md5.

pkg/postgresql/postgresql.go
emdem: @sgotti I gave it a lot of thought and decided to simply force the user to either utilize md5 + password and then over-ride specific connections via the spec/stolonctl update/init.

Need to confirm, but if I recall correctly, it's possible with this code to set a password/md5 auth for super user from the keeper, then to utilize stolonctl or the json spec provided to sentinel to add trust for localhost superuser connections.

I'll start working on the documentation changes when the code changes are at a state you feel comfortable.

emdem:
Added https://github.com/sorintlab/stolon/issues/42 code changes as well. Simply fails with error message indicating version 9.4 or greater is needed.

sgotti:
@emdem Thanks! I thinks this needs some tests but we can make them in another PR. So it looks good to me after removing the unrelated changes:

> Added #42 code changes as well. Simply fails with error message indicating version 9.4 or greater is needed.

Please put this in another pull request since they are two completely different changes.

emdem:
@sgotti https://github.com/sorintlab/stolon/pull/384

seperated the code to a seperate PR

pkg/postgresql/postgresql.go
```diff
@@ -238,8 +242,8 @@ func (p *Manager) Start() error {
 }
 
 func (p *Manager) start(args ...string) error {
-	log.Infow("starting database")
 	name := filepath.Join(p.pgBinPath, "pg_ctl")
+	log.Infow("starting database")
```
sgotti: Why this line has been moved?

pkg/postgresql/postgresql.go
```diff
@@ -360,15 +364,15 @@ func (p *Manager) SetupRoles() error {
 	ctx, cancel := context.WithTimeout(context.Background(), p.requestTimeout)
 	defer cancel()
 
-	if p.suUsername == p.replUsername {
+	if p.suUsername == p.replUsername && p.suAuthMethod != "trust" && p.replAuthMethod != "trust" {
```
sgotti: We should add the replication role to superuser also when authmethod is trust. What shouldn't be done is to set a password. Probably alterRole method should be improved/changed to work with empty passwords.

The same also for the call to `createRole` below (in `creating replication role`).

emdem:
@sgotti I made some changes... I'm testing them a bit more thoroughly. So far... much better.

cmd/keeper/keeper.go
```diff
@@ -1610,37 +1633,57 @@ func keeper(cmd *cobra.Command, args []string) {
...
+		if cfg.pgReplAuthMethod != cfg.pgSUAuthMethod {
+			die("do not support different auth methods when utilizing superuser for replication.")
+		}
+		if cfg.pgSUPassword != cfg.pgReplPassword && cfg.pgSUAuthMethod != "trust" && cfg.pgReplAuthMethod != "trust" {
 			die("provided superuser name and replication user name are the same but provided passwords are different")
 		}
 	}
 
-	if cfg.pgReplPasswordFile != "" {
+	if cfg.pgReplAuthMethod != "trust" && cfg.pgReplPasswordFile != "" {
```
sgotti: The check for `cfg.pgReplAuthMethod != "trust"` can be omitted since it was already checked previously

cmd/keeper/keeper.go
```diff
@@ -1610,37 +1633,57 @@ func keeper(cmd *cobra.Command, args []string) {
...
 		cfg.pgReplPassword, err = readPasswordFromFile(cfg.pgReplPasswordFile)
 		if err != nil {
 			die("cannot read pg replication user password: %v", err)
 		}
 	}
-	if cfg.pgSUPasswordFile != "" {
+	if cfg.pgSUAuthMethod != "trust" && cfg.pgSUPasswordFile != "" {
```
sgotti: The check for `cfg.pgSuAuthMethod != "trust"` can be omitted since it was already checked previously

sgotti:
@emdem Thanks a lot! Merged.

Issue #330, opened 2017-08-01
Deactivate synchronous replication for cluster consisting of 2 nodes in case the slave fails


kimss:
Hi Sorintlab,
please consider the enhancement:
In 2 nodes cluster it makes sense to deactivate the synchronous replication from the master to slave in case the slave fails (or network is splitted) and restore it when the slave backs online. 
Seems it should be the option.

Basic ideas:
1) Though the slave fails the master still contains the correct data and database clients are connected to the master.  From their point of view the database is still online.

sgotti:
@kimss If i got this right you're asking for a "best effort" synchronous replication. Use sync repl if possible but downgrade to async repl if not possible.

As a note, the current sync repl logic is a strict one because the main use of sync repl is to avoid losing data and transactions when electing a new master. So ifsomeone else is going to implement this it should be an option disabled by default.

kimss:
@sgotti 
Yes, correct. I need the synchronous replication will be deactivated in the 2 DB nodes environment if the slave DB disappears.

Would it be possible to implement such a feature?

sgotti:
@kimss it should be possible, it requires some changes in the sentinel code and to add/change the related tests. I'll keep it disabled by default and add a cluster spec option (like 'strictSyncRepl') which defaults to true. I cannot provide an estimate on when it'll be implemented, if no one is going to implement it I can do it after other older opened issues are implemented.

kimss:
@sgotti this is very positive!
Probably it sounds not good, if we disable sync replication in 2 nodes cluster in a moment of master or slave failure we will face the risk of the absent data redundancy. This is not good from the data safety perspective, but from the service perspective it is normal => DB clients will continue their activities.

In out case at this moment infrastructure guys will immediately start working (probably they will be waken up). This reaction might take ~30 minutes. But the applications will continue operate smoothly.

Would it be possible to disable the sync replication not only in case when slave fails, but also if master failes then promote the slave into master with sync replication disabled?

sgotti:
@kimss I'm not sure if I understand your statements but your proposal will obviously trade consistency for availability (in the CAP theorem). For this reason I'll keep this option disabled by default.

If you want consistency you should use sync repl with at least 3 nodes (and 2 sync repl standbys) to be available with one node loss.

[...]

PR #410, opened 2018-01-15
added: sync standby name, additional named replication slot

saranhada:
This pull request relates to the stream https://github.com/sorintlab/stolon/issues/330
The ideas of the 2 new parameters:

1) Implemented "AdditionalSyncStandbyNames" parameter;
   - configure additional value for the "synchronous_standby_names" parameter
2) Implemented "AdditionalReplicationSlotName";
   - provide additional replication slot with specified configurable name 
  

The additional synchronous name will be used in 2 nodes synchronous stolon cluster. If the master or standby fails then postgres will start using additional synchronous name for the synchronous replication, and therefore the stolon cluster (now it consists of the one survived single node) will continue operate. 
As the value for "AdditionalSyncStandbyNames" parameter we want to use PgBarman's application name and since it requires exact replication slot we have implemented the "AdditionalReplicationSlotName" parameter
From the consistency point of view the schema seems to be ok because in a moment of a failure the PgBarman will be engaged as the synchronous destination and the data will be saved by the additional synchronous standby (e.g. PgBarman). 

The pull request is the continue of the discussion: https://github.com/sorintlab/stolon/issues/330


Vadim0908:
This patch allows us to use additional replication slots for backup. For example, we can implement two options for working with a Barman.

# Classic Stolon Architecture and Barman Backup

stolon barman sr 3 nodes

Barman offers 2 options for backup and archiving. Traditionally, Barman has supported standard WAL file shipping through PostgreSQL’s archive_command (usually via rsync/SSH). With this method, WAL files are archived only when PostgreSQL switches to a new WAL file. To keep it simple, this normally happens every 16MB worth of data changes.
Barman 1.6.0 introduces streaming of WAL files for PostgreSQL servers 9.2 or higher, as an additional method for transactional log archiving, through pg_receivexlog. WAL streaming is able to reduce the risk of data loss, bringing RPO down to near zero values.
Barman 2.0 introduces support for replication slots with PostgreSQL servers 9.4 or above, therefore allowing WAL streaming-only configurations. Moreover, you can now add Barman as a synchronous WAL receiver in your PostgreSQL 9.5 (or higher) cluster, and achieve zero data loss (RPO=0).
Thus, the traditional archiving of WAL logs can lead to data loss in the event of an accident. Therefore, we believe that we need to use both approaches for archiving WA-logs.

## Synchronous WAL streaming
This feature is available only from PostgreSQL 9.5 and above. Barman can also reduce the Recovery Point Objective to zero, by collecting the transaction WAL files like a synchronous standby server would.

# Stolon Architecture with 2 Nodes and Barman as 3rd Nodes
stolon barman sr 2 nodes

cmd/keeper/keeper.go
```diff
@@ -796,6 +802,36 @@ func (p *PostgresKeeper) isDifferentTimelineBranch(followedDB *cluster.DB, pgSta
...
+
+	if backupSlotDetected {
+		p.pgm.DropReplicationSlot(backupSlotName)
+	}
+
+	if err = p.pgm.CreateReplicationSlot(backupSlotName); err != nil {
+		log.Errorw("failed to create replication slot", "slotName", backupSlotName, zap.Error(err))
+		return err
+	}
+
```
sgotti: won't this drop and recreate the replication slot at every loop?

cmd/keeper/keeper.go
```diff
@@ -278,14 +278,20 @@ func (p *PostgresKeeper) createPGParameters(db *cluster.DB) common.Parameters {
...
 		// TODO(sgotti) Find a way to detect the pg major version also
 		// when the instance is empty. Parsing the postgres --version
 		// is not always reliable (for example we can get `10devel` for
 		// development version).
 		// Now this will lead to problems starting the instance (or
 		// warning reloading the parameters) with postgresql < 9.6
-		if len(synchronousStandbys) > 1 {
-			parameters["synchronous_standby_names"] = fmt.Sprintf("%d (%s)", len(synchronousStandbys), strings.Join(synchronousStandbys, ","))
+		if syncStandbyNumber > 1 {
```
sgotti: Probably here there's some logic that needs to be detailed in a comment. If I got it right you want to define the additionalsyncstandbynames only if there's at least one defined synchronous standby?

I won't do all of this here. This is something that should be handled by the sentinel. So, inside the sentinel I'll just add the additional synchronous standby in db.Spec.SynchronousStandby.

doc/cluster_spec.md
```diff
@@ -24,7 +24,9 @@ Some options in a running cluster specification can be changed to update the des
...
+| additionalSyncStandbyNames| additional names for possible standbys not included into stolon cluster, but used as replication destinations                                                                                                                                                                                                                                                                                                                                                                     | mo                        | string            | null                                                                                                                                |
```
sgotti: Will do this an []string or you'll have to explicitly define that they should be comma separated and parse/validate them.

pkg/cluster/cluster.go
```diff
@@ -345,6 +356,12 @@ func (os *ClusterSpec) WithDefaults() *ClusterSpec {
 	if s.AdditionalWalSenders == nil {
 		s.AdditionalWalSenders = Uint16P(DefaultAdditionalWalSenders)
 	}
+	if s.AdditionalSyncStandbyNames == nil {
+		s.AdditionalSyncStandbyNames = StringP(DefaultSyncStandbyNames)
+	}
+	if s.AdditionalReplicationSlotName == nil {
+		s.AdditionalReplicationSlotName = StringP(DefaultReplicationSlotName)
+	}
```
sgotti: After making them a `[]string` the values should be validated to check that they're valid sync standby names and replication slot names.

cmd/sentinel/sentinel.go
```diff
@@ -360,6 +360,8 @@ func (s *Sentinel) setDBSpecFromClusterSpec(cd *cluster.ClusterData) {
 			db.Spec.FollowConfig.StandbySettings = clusterSpec.StandbySettings
 		}
 		db.Spec.AdditionalWalSenders = *clusterSpec.AdditionalWalSenders
+		db.Spec.AdditionalSyncStandbyNames = *clusterSpec.AdditionalSyncStandbyNames
```
sgotti: I'll remove this and just add the additional synchronous standbys in db.Spec.SynchronousStandby.

doc/cluster_spec.md
```diff
@@ -24,7 +24,9 @@ Some options in a running cluster specification can be changed to update the des
...
+| additionalReplicationSlotName | name of the additional replication slot which is always created by the master postgres instance                                                                                                                                                                                                                                                                                                                                                                               | no                        | string            | null                                                                                                                                |
```
sgotti: Cannot be this a `[]string` as above so users can define multiple additional replication slots?

pkg/cluster/cluster.go
saranhada: AdditionalReplicationSlotName - is the name of a single replication slot, not an array

sgotti: yes, I was asking to make it an array so multiple additional replication slots can be defined like done for additionalSyncStandbyNames.

saranhada: Simon, in our task we need one definite replication slot. But if you insist we can extend the basic idea. Do you?

sgotti: @saranhada yes, since we are providing a new option to define an additional replication slotlets make it useful for everyone giving the ability to define N additional replication slots if needed.

pkg/cluster/cluster.go
```diff
@@ -530,6 +586,9 @@ type DBSpec struct {
 	// AdditionalWalSenders defines the number of additional wal_senders in
 	// addition to the ones internally defined by stolon
 	AdditionalWalSenders uint16 `json:"additionalWalSenders"`
+
+	AdditionalSyncStandbyNames     string   `json:"additionalSyncStandbyNames"`
```
sgotti: Make this an `[]string`.

pkg/cluster/cluster.go
```diff
@@ -222,9 +229,14 @@ type ClusterSpec struct {
 	// MaxSynchronousStandbys is the maximum number if synchronous standbys
 	// to be configured when SynchronousReplication is true
 	MaxSynchronousStandbys *uint16 `json:"maxSynchronousStandbys,omitempty"`
+
 	// AdditionalWalSenders defines the number of additional wal_senders in
 	// addition to the ones internally defined by stolon
 	AdditionalWalSenders *uint16 `json:"additionalWalSenders"`
+
+	AdditionalSyncStandbyNames     *string `json:"additionalSyncStandbyNames"`
```
sgotti: Make this an `[]string`.

pkg/cluster/cluster.go
```diff
@@ -222,9 +229,14 @@ type ClusterSpec struct {
...
+	AdditionalReplicationSlotNames *string `json:"additionalReplicationSlotNames"`
```
sgotti: Make this an `[]string`.

cmd/keeper/keeper.go
```diff
@@ -278,14 +278,29 @@ func (p *PostgresKeeper) createPGParameters(db *cluster.DB) common.Parameters {
...
+		// If two nodes stolon cluster fails then the applications specified in AdditionalSyncStandbyNames
+		// will be engaged for the synchronous replication.
+		// The "syncStandbyNumber" variable should not be affected here because otherwise the logic will require synchronous
+		// commits for all standby names
+		if len(db.Spec.AdditionalSyncStandbyNames) > 0 {
+			for _, syncStandByName := range db.Spec.AdditionalSyncStandbyNames {
+				synchronousStandbys = append(synchronousStandbys, syncStandByName)
+			}
+		}
+
```
sgotti: I honestly can't understand the logic here. As I tried to explain in my previous comment: https://github.com/sorintlab/stolon/pull/410/files#r161692823 just drop this and let the sentinel add the AdditionalSyncStandbyNames to the `db.Spec.SynchronousStandby`.

cmd/sentinel/sentinel.go
```diff
@@ -360,6 +360,8 @@ func (s *Sentinel) setDBSpecFromClusterSpec(cd *cluster.ClusterData) {
 			db.Spec.FollowConfig.StandbySettings = clusterSpec.StandbySettings
 		}
 		db.Spec.AdditionalWalSenders = *clusterSpec.AdditionalWalSenders
+		db.Spec.AdditionalSyncStandbyNames = clusterSpec.AdditionalSyncStandbyNames
```
sgotti: Don't define `db.Spec.AdditionalSyncStandbyNames` but just add the additional synchronous standbys to db.Spec.SynchronousStandby.

saranhada: @sgotti We want these additional standbys not to be taken into account here keeper.go:303

			parameters["synchronous_standby_names"] = fmt.Sprintf("%d (%s)", syncStandbyNumber, strings.Join(synchronousStandbys, ","))
		} else {
			parameters["synchronous_standby_names"] = strings.Join(synchronousStandbys, ",")
		}

sgotti: @saranhada I don't understand why. As the name states they are `AdditionalSyncStandbyNames` so they should be added to the list of "synchronous_standby_names" no?

sgotti: @saranhada If you want them only as a "fallback" standby when one or more keeper standby isn't working then that's another story and it should be handled differently since this will break the current concistency logic that uses a fake standby as this change will skip it.

As I explained [previously]( https://github.com/sorintlab/stolon/pull/410#pullrequestreview-89021195) these are 2 distinct changes. One is to set additional sync standbys (not replacement standbys) that will be counted in the number of requires sync standbys. Another is to disable sync repl when you don't have enough standbys.

saranhada: @sgotti if I add the additional standby names into db.Spec.SynchronousStandby then the provided piece of will be generated like (for instance) = "2 (stolon_fdg4gs, add_name)", and this is what I want to avoid. We want to implement additional synchronous standbys for needs of backups as it was mentioned on the schema (before). It is supposed that PgBarman will serve the required needs. Thus we add PgBarman into 2 nodes stolon cluster with enabled sync replication. In a moment the master fails the standby keeper will be promoted into new Master with sync replication still enabled. The single stolon node will still operate and the PgBarman will be used as sync standby destination.

Later when a new 2nd node of stolon is restored/created the cluster will be switched to use it. But PgBarman will serve the needs of sync backup.

sgotti: @saranhada Looks like you have two requirements:

a. You don't want that a primary blocks when there are less than `MinSynchronousStandbys` working standbys AND **you don't want to just fallback to async repl**  but instead you want to send wal records to an external standby that will become Synchronous ONLY when the above condition happen.

Note that this isn't #330 (deactivate synchronous replication for cluster consisting of 2 nodes in case the slave fails).

and

b. You want an external synchronous standby for backups with rpo = 0 but only when there are less than `MinSynchronousStandbys` working standbys.

But I'm quite sure they can be summed in only one requirement:

* Add an option to define N **Fallback** synchronous standbys that will be added only when there are less than `MinSynchronousStandbys` working standbys.

The main issue with your implementation is that you can't control all of this using the FIRST method of the `sync_standby_names` parameters because it's acting outside stolon control (switching sync stanbys without the sentinel knowing this). Let me explain:

This PR will define for the primary instance something like this: `1 (stolon_fdg4gs, add_name)`.

This means that if the standby keeper `stolon_fdg4gs` fails, the master instead of blocking will continue committing transactions since it can send data to pgBarman (or any other external pg instance defined by `add_name`).

If in a small time window a standby dies then the primary dies and the failed standby comes back before the sentinel notices this, it may happen that the standby is elected as master also if it's not in sync with the dead primary causing the loss of all the changes contained in the missed wal records.

That's the reason why we are forced to have FIRST value equal to the total number sync standbys. (BTW this check https://github.com/sorintlab/stolon/pull/410/files#diff-041abc9a078974534612e6fc3690c2bbL287we is not needed and we should just use this syntax: https://github.com/sorintlab/stolon/pull/410/files#diff-041abc9a078974534612e6fc3690c2bbL290 (not sure why I used two formats that means the same). Will open a PR to remove this confusion and document it better).

That's the reason why I was asking to just add the `AdditionalSyncStandbys` to `db.Spec.SynchronousStandbys`. But obviously this will block your primary if just one of the external sync standbys are down and you don't want this.

## Proposed implementation

So the starting point is that it must be the sentinel to control all of this. And let's change the option name to some more clear like `ExternalFallbackSyncStandbys`.

A possible solution is to handle all of this in the sentinel (like we already do adding a fakestandby when there aren't enough keeper to force the primary to block): if the sentinel detects that the good standbys are less than `MinSynchronousStandbys`  then add one or more `ExternalFallbackSyncStandbys` to reach `MinSynchronousStandbys`.

So, in a two node cluster, you'll have `MinSynchronousStandbys = 1`, if the standbys dies the sentinel will notice this, remove it from the `db.Spec.SynchronousStandbys` and add the first entry in `ExternalFallbackSyncStandbys`. In this way the primary will unblock. The above problem won't happen because the external standby will be added by the sentinel only if it detects that the internal standbys is dead. If the dead standby comes back it'll be readded to the list of sync stanbys and the external stanby will be removed.

Of course the switch isn't instantaneus as if it's managed by the postgres instance but will permit the cluster to unblock after some seconds/minutes.

sgotti:
@saranhada Looks like you have two requirements:

a. You don't want that a primary blocks when there are less than `MinSynchronousStandbys` working standbys AND **you don't want to just fallback to async repl**  but instead you want to send wal records to an external standby that will become Synchronous ONLY when the above condition happen.

Note that this isn't #330 (deactivate synchronous replication for cluster consisting of 2 nodes in case the slave fails).

and

b. You want an external synchronous standby for backups with rpo = 0 but only when there are less than `MinSynchronousStandbys` working standbys.

But I'm quite sure they can be summed in only one requirement:

* Add an option to define N **Fallback** synchronous standbys that will be added only when there are less than `MinSynchronousStandbys` working standbys.

The main issue with your implementation is that you can't control all of this using the FIRST method of the `sync_standby_names` parameters because it's acting outside stolon control (switching sync stanbys without the sentinel knowing this). Let me explain:

This PR will define for the primary instance something like this: `1 (stolon_fdg4gs, add_name)`.

This means that if the standby keeper `stolon_fdg4gs` fails, the master instead of blocking will continue committing transactions since it can send data to pgBarman (or any other external pg instance defined by `add_name`).

If in a small time window a standby dies then the primary dies and the failed standby comes back before the sentinel notices this, it may happen that the standby is elected as master also if it's not in sync with the dead primary causing the loss of all the changes contained in the missed wal records.

That's the reason why we are forced to have FIRST value equal to the total number sync standbys. (BTW this check https://github.com/sorintlab/stolon/pull/410/files#diff-041abc9a078974534612e6fc3690c2bbL287we is not needed and we should just use this syntax: https://github.com/sorintlab/stolon/pull/410/files#diff-041abc9a078974534612e6fc3690c2bbL290 (not sure why I used two formats that means the same). Will open a PR to remove this confusion and document it better).

That's the reason why I was asking to just add the `AdditionalSyncStandbys` to `db.Spec.SynchronousStandbys`. But obviously this will block your primary if just one of the external sync standbys are down and you don't want this.

## Proposed implementation

So the starting point is that it must be the sentinel to control all of this. And let's change the option name to some more clear like `ExternalFallbackSyncStandbys`.

A possible solution is to handle all of this in the sentinel (like we already do adding a fakestandby when there aren't enough keeper to force the primary to block): if the sentinel detects that the good standbys are less than `MinSynchronousStandbys`  then add one or more `ExternalFallbackSyncStandbys` to reach `MinSynchronousStandbys`.

So, in a two node cluster, you'll have `MinSynchronousStandbys = 1`, if the standbys dies the sentinel will notice this, remove it from the `db.Spec.SynchronousStandbys` and add the first entry in `ExternalFallbackSyncStandbys`. In this way the primary will unblock. The above problem won't happen because the external standby will be added by the sentinel only if it detects that the internal standbys is dead. If the dead standby comes back it'll be readded to the list of sync stanbys and the external stanby will be removed.

Of course the switch isn't instantaneus as if it's managed by the postgres instance but will permit the cluster to unblock after some seconds/minutes.

saranhada:
@sgotti I'm agree the name ExternalFallbackSyncStandbys sounds good. It really reflects the idea of the backup synchronous standbys.

Also I have reviewed the stolon source code again, of course I'm not an deep expert in it. 

You proposed to migrate the logic into sentinel. I think I see a piece of code can be responsible for the creation of the required configuration. It seems to start here sentinel.go:952 => 

if !masterOK {
      log.Infow("trying to find a new master to replace failed master")
.....
newMasterDB.Spec.SynchronousStandbys = []string{fakeStandbyName}


I have the following proposal:
 - If the ExternalFallbackSyncStandbys parameter is not empty then just take its values and put into the "newMasterDB.Spec.SynchronousStandbys" list. 

How do you think?

sgotti:
@saranhada Without testing a start to see if it works or there're some corner cases could be:

* Before https://github.com/sorintlab/stolon/blob/70464d31728f543b38d7c21ad09c28398105b8ab/cmd/sentinel/sentinel.go#L1194 add a check `if len(synchronousStandbys) < int(*clusterSpec.MinSynchronousStandbys) { ` that will add N `ExternalFallbackSyncStandbys` to to reach `MinSynchronousStandbys`.

* Validate in `clusterspec.Validate()` that the provided `ExternalFallbackSyncStandbys` are not stolon names (use `IsStolonName` in https://github.com/sorintlab/stolon/blob/70464d31728f543b38d7c21ad09c28398105b8ab/common/common.go#L65) to not clash with internal stolon names.

* I'll add some tests to see that all of this works to both `sentinel/sentinel_test.go` and integration tests in `tests/integration`

saranhada:
@sgotti 
Your suggestion:
_add a check if len(synchronousStandbys) < int(*clusterSpec.MinSynchronousStandbys) { that will add N ExternalFallbackSyncStandbys to to reach MinSynchronousStandbys._
will not work.


Please see the code in keeper.go:275:


`// Setup synchronous replication`
`if db.Spec.SynchronousReplication && len(db.Spec.SynchronousStandbys) > 0 {`
`		synchronousStandbys := []string{}`
`		for _, synchronousStandby := range db.Spec.SynchronousStandbys {`
`			synchronousStandbys = append(synchronousStandbys, common.StolonName(synchronousStandby))`
`		}`

Every sync name which we add in sentinel will be prefixed by "stolon_" in keeper.

I think we should implement the mentioned by you logic (add required the insufficient number of sync standby names) in keeper and take into consideration the MinSynchronousStandbys & MaxSynchronousStandbys

sgotti:
> Every sync name which we add in sentinel will be prefixed by "stolon_" in keeper.

So lets remove this from the keeper and make the sentinel to generate a `StolonName`. The keeper will put the provided synchronous standby names as provided in the db spec without any change.

Also remove all the prefix checks and removal in the keeper: https://github.com/sorintlab/stolon/blob/90a16f5c494600ee92251393545fc07900d4ba92/cmd/keeper/keeper.go#L595

to report them as is (so also the external stanbys are reported).

Perhaps there's something else to change in the sentinel. Let's see after these changes.


> I think we should implement the mentioned by you logic (add required the insufficient number of sync standby names) in keeper and take into consideration the MinSynchronousStandbys & MaxSynchronousStandbys

The keeper just tries to converge to the clusterdata db spec and for this reason must not implement any cluster logic. The clusterdata db spec is calculated by the sentinel so all of this logic pertains to the sentinel.

sgotti:
@saranhada I rebased you branch on master and changed some parts to make it work. You can find it here (just one commit): https://github.com/sgotti/stolon/tree/saranhada-ADD_PARAMS_SYNC

* In the end, instead of keeping both internal and external standbys in `db.Spec.SynchronousStanbys`, I preferred to add a `db.Spec.ExternalSynchronousStandbys` since these are two different things. Se below.
* Added some comments to clarify that `db.Spec.SynchronousStanbys` is a list of internal db UIDs (not generic standby names) while `db.Spec.ExternalSynchronousStandbys` is a list of external standby names.
* For the above reason the `fakestandbyname` will now be added to `db.Spec.ExternalSynchronousStandbys`

Please take a look. 

You can just take my commit and put it in your branch and force push to update this PR (no need to open a new PR).

marct83:
This is completely unrelated to this issue....but does Barman work with Stolon and is configuration any different than a regular Barman Postgres setup?

sgotti:
> This is completely unrelated to this issue....but does Barman work with Stolon and is configuration any different than a regular Barman Postgres setup?

@marct83 can you please ask this question on the gitter channel so more people can help you and we don't pollute this PR? Thanks.

marct83:
No problem. Delete my comments. Thanks

sgotti:
@saranhada I saw you imported my commit but also added another commit (https://github.com/sorintlab/stolon/pull/410/commits/2f25a26f3eb15e0976c95789308e0459ef6d2ce1) that is adding unneeded things (`ExternalFallbackSyncStandbys` is not used and need).

You said on gitter that you completed manual tests but before this PR can be merged we need some automatic tests. We need some unit tests in `cmd/sentinel/sentinel.go` and an integration tests in `tests/integration/ha_test.go`. For this last one a possible implementation idea is to use the Standby cluster feature to set up a stanby cluster as a synchronous standby (it'll be the same as you barman instance but we can reuse the code used in other tests) and do all the thing you probably did manually.

If you have some problems adding them I can do it for you in the next days when I have some time.

saranhada:
@sgotti sorry for the late response.
Can you please confirm now that the created code is OK?  ...and I should created auto-tests.
If not then I'd prefer to accomplish with the code/logic and after that to create tests. I'm absolutely agree that I have to create the tests as well.

pkg/cluster/cluster.go
```diff
@@ -397,6 +401,45 @@ func (os *ClusterSpec) Validate() error {
...
+				return fmt.Errorf("External fallback synchronous standby name \"%s\" has an incorrect format, it should not have prefix 'stolon_'", standbyName)
+			}
+			if strings.Contains(standbyName, "\n") {
+
+			}
+		}
+	}
+
+	if len(s.AdditionalReplicationSlotNames) > 0 {
+		rp := regexp.MustCompile("[_0-9a-z]+")
```
sgotti: Can't you just use https://github.com/sorintlab/stolon/blob/master/pkg/postgresql/utils.go#L328 ?

pkg/cluster/cluster.go
```diff
@@ -397,6 +401,45 @@ func (os *ClusterSpec) Validate() error {
 	if s.InitMode == nil {
 		return fmt.Errorf("initMode undefined")
 	}
+
+	symbolsNameFilter := regexp.MustCompile("[_0-9]")
+	if len(s.ExternalFallbackSyncStandbys) > 0 {
+		correctSymbolsNameFilter := regexp.MustCompile("[_0-9A-Za-z]")
+		for _, standbyName := range s.ExternalFallbackSyncStandbys {
```
sgotti: Can you put this in its own function and add some unit tests to it?

sgotti:
@saranhada The logic related to the `FallbackExternalSyncStandbys` looks good but I'm quite sure there'll be bugs that we could catch only if we write tests to cover the changes. As I explained before I don't understand why you added commit https://github.com/sorintlab/stolon/commit/2f25a26f3eb15e0976c95789308e0459ef6d2ce1

Additionally:

* The part related to creating additional replication slot is a different feature than the `FallbackExternalSyncStanbys` and should be split in its own PR. I haven't reviewed it because it's difficutl to see if it does what it's meant to do without some tests. Can you open a new PR for it so I can review it separately from this?

* Added some comment for the new fields validation.

saranhada:
@sgotti do you suggest to split the PR into the following 2?
1) Additional replication slots
2) Fallback sync standbys

Regarding option 2 I understand the integration tests are required. But how about option 1? DO you require it? I think unit tests will be enough.

sgotti:
> @sgotti do you suggest to split the PR into the following 2?
> 
> 1. Additional replication slots
> 2. Fallback sync standbys

yes.

> Regarding option 2 I understand the integration tests are required. But how about option 1? DO you require it? I think unit tests will be enough.

Now the current `updateReplSlots` is covered by the current integration tests. In its current shape we cannot cover it with just an unit test since to test it we need a running keeper and postgres instance. To make an unit test you should split it in two part, one that does the get and the drop/create on the postgres instance and a function than given the current replication slots and the desired slots will return a list of slots to drop and a list of slots to create (so an unit test can be created on that function).

But I'd say to just start splitting it in its own PR, then let's see which kind of tests are needed. I could write something myself to let you see how this test could be done.

saranhada:
Reopened

cmd/keeper/cmd/keeper.go
```diff
@@ -312,6 +312,7 @@ func (p *PostgresKeeper) createPGParameters(db *cluster.DB) common.Parameters {
 		for _, synchronousStandby := range db.Spec.SynchronousStandbys {
 			synchronousStandbys = append(synchronousStandbys, common.StolonName(synchronousStandby))
 		}
+		numberSyncStandbys := len(synchronousStandbys)
```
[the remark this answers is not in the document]
saranhada: If not do it then then the following code will result in forming of incorrect "synchronous_standby_names", it will will be like "2 (standby1, standby2)": 
		for _, synchronousStandby := range db.Spec.ExternalSynchronousStandbys {
			synchronousStandbys = append(synchronousStandbys, synchronousStandby)
		}
By saving the size of synchronousStandbys we are avoiding the for compilation of the incorrect value

cmd/keeper/cmd/keeper.go
```diff
@@ -331,7 +332,7 @@ func (p *PostgresKeeper) createPGParameters(db *cluster.DB) common.Parameters {
 		// which of the 3 standbys is really synchronous and in sync with the
 		// master. And choosing the non synchronous one will cause the loss of
 		// the transactions contained in the wal records not transmitted.
-		if len(synchronousStandbys) > 1 {
+		if numberSyncStandbys > 1 {
```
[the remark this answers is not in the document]
saranhada: Please see the previous comment

cmd/keeper/cmd/keeper.go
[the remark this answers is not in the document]
sgotti: `2 (standby1, standby2)` is correct. As the comment below in the code says all the provided synchronous standbys must be synchronous. This is what I explained previously here: https://github.com/sorintlab/stolon/pull/410#issuecomment-360140604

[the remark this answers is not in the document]
saranhada: @sgotti I'm just wanted to notify you that we are currently considering your point of view. Will return to you in a couple of days....

sgotti:
> May I ask you to point a piece code which is responsible for the specified scenario?

Perhaps I lost something but wasn't the purpose of this patch to do this? And this should be done here: https://github.com/sorintlab/stolon/pull/410/files#diff-9961d4546ef09e0152c558e820699523R1324 

When the sentinel declares the standby as not good it'll catch this code and add an external synchronous standby.

sgotti:
@saranhada yes

PR #412, opened 2018-01-15
Feature dynamic hba

t0k4rt:
Hello, 
We want to deploy Stolon in our architecture, but we noticed that connection to su and repl users are open to the world which is (imho) not good for a secure cluster. 
We can deal with this with a firewall rule but I would prefer this to be setup correctly in the pg_hba.conf.
I saw a todo in your code concerning this issue so I decide to propose a fix. I'm not goloang developer nor sure it will match your coding standard or code organization wishes ... but it's a start.

The code simply read the cluster data generate a list of pghba entries and merges it with the `db.Spec.PGHBA`.

I found this to be the easiest solution.

Important to notice, it does not compile, I've got the current error: 
```
# github.com/sorintlab/stolon/vendor/github.com/sorintlab/tcpkeepalive
gopath/src/github.com/sorintlab/stolon/vendor/github.com/sorintlab/tcpkeepalive/tcpkeepalive.go:14:10: undefined: setIdle
gopath/src/github.com/sorintlab/stolon/vendor/github.com/sorintlab/tcpkeepalive/tcpkeepalive.go:22:10: undefined: setCount
gopath/src/github.com/sorintlab/stolon/vendor/github.com/sorintlab/tcpkeepalive/tcpkeepalive.go:29:10: undefined: setInterval
```

I also changed the default su and repl connection to localhost.

I hope it can be useful for you ! 
Don't hesitate to tell me how I can improve this.
Thanks for this product !
Alexandre


t0k4rt:
./test does not launch on OSX either

cmd/keeper/keeper.go
```diff
@@ -1512,7 +1513,8 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 
 	if !reflect.DeepEqual(db.Spec.PGHBA, pgm.CurHba()) {
 		log.Infow("postgres hba entries changed, reloading postgres instance")
-		pgm.SetHba(db.Spec.PGHBA)
+		// Dynamicly generate hba auth from clusterData
+		pgm.SetHba(append(db.Spec.PGHBA, p.generateHBA(cd)...))
```
sgotti: The generated hba entries should be inserted before the user provided hba entries.

sgotti: don't append again here but do the append directly inside `generateHBA`.

t0k4rt:
No Problems, I'll change that !
I'm not experienced at all with go nor go vendors management, so please excuse this poor (yet) PR.

sgotti:
> I'm not experienced at all with go nor go vendors management, so please excuse this poor (yet) PR.

don't worry and welcome to the fantastic world of Go! Feel free to ask for any doubt (here on on gitter).

t0k4rt:
Hi, I updated the code as requested.
I added an optional switch defaultReplAccess that can take two options: "all" (default value) or "standby"

In case user select all:
* master will have its default hba with:
  * repl access to all ips (0.0.0.0/0)
  * su access to all ips (0.0.0.0/0)
* slave will have its default hba with:
  * su access to all ips (0.0.0.0/0)
  * no replication access since slave does not need it

In case user select standby:
* master will have its default hba with:
  * repl access to all standby ips (x.x.x.x/32)
  * su access to localhost (127.0.0.1/32)
  * su access to all standby ips (x.x.x.x/32)
* slave will have its default hba with:
  * su access to localhost (127.0.0.1/32) 
  * no replication access since slave does not need it

Maybe I should seperate repl and su options and create defaultReplAccess / defaultSuAccess ?

cmd/keeper/keeper.go
```diff
@@ -89,6 +89,7 @@ type config struct {
 	pgSUPasswordFile        string
 	pgInitialSUUsername     string
 	pgInitialSUPasswordFile string
+	defaultReplAccess       string
```
sgotti: Sorry, I meant to set it as a cluster spec option, so it could be set globally: https://github.com/sorintlab/stolon/blob/master/pkg/cluster/cluster.go#L186

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1762,51 @@ func keeper(c *cobra.Command, args []string) {
...
+
+	if p.defaultReplAccess == "all" {
+		// allow for all servers su access
+		hbaBase = append(
+			hbaBase,
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "0.0.0.0/0", p.pgSUAuthMethod),
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "::0/0", p.pgSUAuthMethod),
+		)
+		// allow only for master server replication access
+		if role == common.RoleMaster {
```
sgotti: This is a change from the current behaviour and it's breaking the standby cluster feature and related tests. So it should be removed.

cmd/keeper/keeper.go
t0k4rt: Not a problem I'll change that

cmd/keeper/keeper.go
t0k4rt: I don't understand why it breaks. 
Master pg_hba should only allow replication user to connect from standby servers.
Disbaled or enabling replication user on standby server should not have any effects. 
When a standby server is promoted, its pg_hba should be automatically updated and would contain the right rules.

sgotti: Stolon can also be in standby cluster mode. In this case the master keeper manages a standby instance that will sync with an external instance. So the other keepers will sync from it (cascading replication) and the above check will break.

I suggested to just remove it because to keep backward compatibility with "all" mode we have to keep the previous rules.

cmd/keeper/keeper.go
t0k4rt: Changes done

cmd/keeper/keeper.go
t0k4rt: Ok, changes done. It should be ok !

t0k4rt:
I got during tests this issue, So I'm going to do more tests
```
utils.go:121: [keeper c594fd73]: panic: runtime error: invalid memory address or nil pointer dereference
	utils.go:121: [keeper c594fd73]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xadb63f]
```

pkg/cluster/cluster.go
```diff
@@ -183,6 +184,21 @@ type StandbySettings struct {
 	RecoveryMinApplyDelay string `json:"recoveryMinApplyDelay,omitempty"`
 }
 
+type PgHBAAccessMode string
+
+const (
+	// Allow hba access to all members (0.0.0.0/0)
+	PgHBAAccessAll PgHBAAccessMode = "all"
+	// Allow hba access to standby server only
+	PgHBAAccessStandby PgHBAAccessMode = "standby"
```
sgotti: Stale field?

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1756,45 @@ func keeper(c *cobra.Command, args []string) {
 
 	lockFile.Close()
 }
+
+func (p *PostgresKeeper) generateHBA(cd *cluster.ClusterData, role common.Role, userHba []string) []string {
+
+	// Dynamically generate hba auth from clusterData
+	var hbaBase []string
+
+	if cluster.PgHBAAccessAll == *cd.Cluster.Spec.PgHBAAccess {
```
sgotti: Use cd.Cluster.DefSpec().PgHBAAccess or you'll miss the default values when not provided

sgotti: use a switch statement here?

pkg/cluster/cluster.go
```diff
@@ -243,6 +259,10 @@ type ClusterSpec struct {
 	ExistingConfig *ExistingConfig `json:"existingConfig,omitempty"`
 	// Standby setting when role is standby
 	StandbySettings *StandbySettings `json:"standbySettings,omitempty"`
+	// Set default hba auth for replication user and superuser.
+	// Values can be "all" or "standby", "all" allow access to all ips, "standby" restrict master access to standby servers ips.
```
sgotti: standby or strict?

pkg/cluster/cluster.go
```diff
@@ -243,6 +259,10 @@ type ClusterSpec struct {
...
+	// default is "all"
+	PgHBAAccess *PgHBAAccessMode `json:"pgHBAAccess,omitempty"`
```
sgotti: I'll call this `DefaultSUReplAccessMode`, longer but a bit clearer. Other Suggestions?

cmd/keeper/keeper.go
t0k4rt: Done !

t0k4rt: Done !

pkg/cluster/cluster.go
t0k4rt: IMHO it's better strict

pkg/cluster/cluster.go
t0k4rt: We could use:
* DefaultSUReplAccessMode as option
* SUReplAccessMode as type
* DefaultSUReplAccess as default value
* SUReplAccessAll and SUReplAccessStrict as option names 
Is it ok for you ?

sgotti: Seems Good.

cmd/keeper/keeper.go
```diff
@@ -1510,9 +1511,11 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 		log.Infow("postgres parameters not changed")
 	}
 
-	if !reflect.DeepEqual(db.Spec.PGHBA, pgm.CurHba()) {
+	// Dynamicly generate hba auth from clusterData
+	new_hba := p.generateHBA(cd, localRole, db.Spec.PGHBA)
```
sgotti: s/new_hba/newHBA/

t0k4rt:
I've got an issue when I try to update my ClusterSpec, 
the key "defaultSUReplAccessMode": "strict" is properly set but not persisted. 
Any ideas ?

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,45 @@ func keeper(c *cobra.Command, args []string) {
...
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "0.0.0.0/0", p.pgSUAuthMethod),
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "::0/0", p.pgSUAuthMethod),
+			fmt.Sprintf("host replication %s %s %s\n", p.pgReplUsername, "0.0.0.0/0", p.pgReplAuthMethod),
+			fmt.Sprintf("host replication %s %s %s\n", p.pgReplUsername, "::0/0", p.pgReplAuthMethod),
+		)
+	case cluster.SUReplAccessStrict:
+		// allow for all servers su access
+		hbaBase = append(
+			hbaBase,
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "127.0.0.1/32", p.pgReplAuthMethod),
```
sgotti: This is only from 127.0.0.1/32 not from all the standbys or am I missing something?

pkg/postgresql/postgresql.go
```diff
@@ -713,21 +713,10 @@ func (p *Manager) writePgHba() error {
...
-	f.WriteString(fmt.Sprintf("host all %s %s %s\n", p.suUsername, "::0/0", p.suAuthMethod))
-
-	// By default accept all replication connections for the replication user with md5 auth
-	// TODO(sgotti) Configure this dynamically based on our followers provided by the clusterview
-	f.WriteString(fmt.Sprintf("host replication %s %s %s\n", p.replUsername, "0.0.0.0/0", p.replAuthMethod))
-	f.WriteString(fmt.Sprintf("host replication %s %s %s\n", p.replUsername, "::0/0", p.replAuthMethod))
-
 	if p.hba != nil {
 		for _, e := range p.hba {
 			f.WriteString(e + "\n")
```
sgotti: You have forgot to handle the "else" case below inside generateHBA. Here it'll always be skipped since p.hba is never empty.

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,45 @@ func keeper(c *cobra.Command, args []string) {
...
+
+	// Dynamically generate hba auth from clusterData
+	var hbaBase []string
+
+	switch *cd.Cluster.DefSpec().DefaultSUReplAccessMode {
+	case cluster.SUReplAccessAll:
+		hbaBase = append(
+			hbaBase,
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "0.0.0.0/0", p.pgSUAuthMethod),
+			fmt.Sprintf("host all %s %s %s\n", p.pgSUUsername, "::0/0", p.pgSUAuthMethod),
```
sgotti: remove `\n` in all the lines or an empty line will be added

sgotti:
> the key "defaultSUReplAccessMode": "strict" is properly set but not persisted.

What do you mean by "not persisted"? I tried a stolonctl patch and it has been applied to the cluster spec and the pghba file has been updated (with some problems, see my comments) by the keepers.

pkg/postgresql/postgresql.go
t0k4rt: Ok, this means that previously, when p.hba was empty, the generated hba was:

```
host replication repluser 0.0.0.0/0 md5
host replication repluser ::0/0 md5
host all superuser 0.0.0.0/0 md5
host all superuser ::0/0 md5
host all all 0.0.0.0/0 md5
host all all ::0/0 md5
```
which is same as this:
```
host all all 0.0.0.0/0 md5
host all all ::0/0 md5
```
but simpler so I think this should be the result in "all" access mode.

In "strict" access mode we should not use this rule.

cmd/keeper/keeper.go
t0k4rt: Done

t0k4rt:
@sgotti Ok the not persisted issue is because I've got a cluster running two different versions of the keeper. The other keeper does not know the new key so it might removes it.

pkg/postgresql/postgresql.go
sgotti: Not always, the first lines can have a different auth method than md5 (currently only trust but in future other methods could be added), while the last lines are added only if there's no user provided pgHBA and always have md5. I'd prefer to not change the current logic.

sgotti:
> @sgotti Ok the not persisted issue is because I've got a cluster running two different versions of the keeper. The other keeper does not know the new key so it might removes it.

The unique component that writes the cluster data is the sentinel (and stolonctl) so perhaps you had an old sentinel.

pkg/postgresql/postgresql.go
t0k4rt: Yes you're right ! 

For the strict mode, I still would prefer not to add this rule.

sgotti: It may be confusing but these are two separate things. The first lines (the one affected by this patch to set the DefaultSuReplAccessMode) are for instance replication, the latter is a default when the user doesn't provide a PGHBA. So I'll leave them distinct and not make DefaultSuReplAccessMode change this.

t0k4rt: Ok I understand !
Can I plan a DefaultUserAccessMode feature ?

pkg/postgresql/postgresql.go
```diff
@@ -713,30 +713,14 @@ func (p *Manager) writePgHba() error {
 
 	// Minimal entries for local normal and replication connections needed by the stolon keeper
 	// Matched local connections are for postgres database and suUsername user with md5 auth
-	// Matched local replicaton connections are for replUsername user with md5 auth
+	// Matched local replication connections are for replUsername user with md5 auth
 	f.WriteString(fmt.Sprintf("local postgres %s %s\n", p.suUsername, p.suAuthMethod))
 	f.WriteString(fmt.Sprintf("local replication %s %s\n", p.replUsername, p.replAuthMethod))
```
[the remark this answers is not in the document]
t0k4rt: Done

pkg/postgresql/postgresql.go
sgotti: @t0k4rt you can just do the same proving an empty (not nil) pgHBA in the clusterspec

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,63 @@ func keeper(c *cobra.Command, args []string) {
...
+	}
+
+	// when no custom hba is set, we default to open to all
+	if len(userHba) == 0 {
+		computedHBA = append(
+			computedHBA,
+			"host all all 0.0.0.0/0 md5",
+			"host all all ::0/0 md5",
+		)
+	}
```
sgotti: this must be at the end. After the stolon generated rules.

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,63 @@ func keeper(c *cobra.Command, args []string) {
...
+	// Minimal entries for local normal and replication connections needed by the stolon keeper
+	// Matched local connections are for postgres database and suUsername user with md5 auth
+	// Matched local replication connections are for replUsername user with md5 auth
+	computedHBA := []string{
+		fmt.Sprintf("local postgres %s %s\n", p.pgSUUsername, p.pgSUAuthMethod),
+		fmt.Sprintf("local replication %s %s\n", p.pgReplUsername, p.pgReplAuthMethod),
+	}
+
+	// when no custom hba is set, we default to open to all
+	if len(userHba) == 0 {
```
sgotti: don't use len but check if it's nil or not like currently done. So you can distinguis between a not provided (nil) or a provided but empty pgHBA.

pkg/cluster/cluster.go
```diff
@@ -243,6 +257,10 @@ type ClusterSpec struct {
 	ExistingConfig *ExistingConfig `json:"existingConfig,omitempty"`
 	// Standby setting when role is standby
 	StandbySettings *StandbySettings `json:"standbySettings,omitempty"`
+	// Set default hba auth for replication user and superuser.
+	// Values can be "all" or "strict", "all" allow access to all ips, "strict" restrict master access to standby servers ips.
```
sgotti: s/to/from/g

pkg/cluster/cluster.go
```diff
@@ -243,6 +257,10 @@ type ClusterSpec struct {
 	ExistingConfig *ExistingConfig `json:"existingConfig,omitempty"`
 	// Standby setting when role is standby
 	StandbySettings *StandbySettings `json:"standbySettings,omitempty"`
+	// Set default hba auth for replication user and superuser.
```
sgotti: `Define the mode of the default hba rules needed for replication by standby keepers (the su and repl auth methods will be the one provided in the keeper command line options)`

cmd/keeper/keeper.go
t0k4rt: Done

cmd/keeper/keeper.go
t0k4rt: Done

pkg/cluster/cluster.go
t0k4rt: Done

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,65 @@ func keeper(c *cobra.Command, args []string) {
...
+	case cluster.SUReplAccessStrict:
+		// allow for all servers su access
+		computedHBA = append(
+			computedHBA,
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "127.0.0.1/32", p.pgReplAuthMethod),
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "::1/128", p.pgReplAuthMethod),
+		)
+		// allow only for master replication access
+		if role == common.RoleMaster {
+			// we should not allow replication acces for 127.0.0.1 since it can be then forwarded through proxy.
```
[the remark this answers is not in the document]
t0k4rt: No problems, I'll check that !
I'll update the cluster spec too
Concerning an integration test I'll check too but not sure I'll managed to do something good enough ...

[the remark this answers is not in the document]
sgotti: > Concerning an integration test I'll check too but not sure I'll managed to do something good enough ...

Don't worry, I could add something later. Unfortunately integration tests of distributed system is quite tedious and complex...

t0k4rt:
@sgotti Hello, everything should be fixed.
Have you any other feedback ? 
thanks !

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,94 @@ func keeper(c *cobra.Command, args []string) {
...
+	// get keeper current db, current db should not be nil
+	k, ok := cd.Keepers[p.keeperLocalState.UID]
+	if !ok {
+		panic(fmt.Sprintf("Unable to get Keeper data with keeper UID %s", p.keeperLocalState.UID))
+	}
+	currentDb := cd.FindDB(k)
+	if currentDb == nil {
+		panic(fmt.Sprintf("Unable to get current keeper local DB with UID %s", p.keeperLocalState.UID))
+	}
+
```
sgotti: * you already have the `db` in the caller just pass it to generateHBA without all of the above (and avoid calls to panic)
* `role common.Role` isn't used anymore
* rename `userHba` to `specHBA` to be clearer.

generateHBA should become:
```
func (p *PostgresKeeper) generateHBA(cd *cluster.ClusterData, db *cluster.DB, specHBA []string) []string {
```

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,94 @@ func keeper(c *cobra.Command, args []string) {
...
+						computedHBA,
+						fmt.Sprintf("host replication %s %s/32 %s", p.pgReplUsername, db.Status.ListenAddress, p.pgReplAuthMethod),
+						fmt.Sprintf("host all %s %s/32 %s", p.pgSUUsername, db.Status.ListenAddress, p.pgReplAuthMethod),
+					)
+				}
+			}
+		}
+	}
+
+	// when no custom hba is set, we default to open to all
```
sgotti: let's keep the original comment, this isn't completely correct.

```
// By default, if no custom pg_hba entries are provided, accept
// connections for all databases and users with md5 auth
```

cmd/keeper/keeper.go
t0k4rt: Ok but since I get specHBA from `db.spec.PGHBA` I could simplify and just only call `func (p *PostgresKeeper) generateHBA(cd *cluster.ClusterData, db *cluster.DB) []string` no ?

sgotti: right!

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,86 @@ func keeper(c *cobra.Command, args []string) {
...
+			fmt.Sprintf("host replication %s %s %s", p.pgReplUsername, "::0/0", p.pgReplAuthMethod),
+		)
+
+	// Case strict access mode
+	case cluster.SUReplAccessStrict:
+		// allow for all servers su access
+		// we should not allow replication acces for 127.0.0.1 since it can be then forwarded through proxy.
+		computedHBA = append(
+			computedHBA,
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "127.0.0.1/32", p.pgSUAuthMethod),
```
sgotti: shouldn't this be `127.0.0.0/8` ?

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,86 @@ func keeper(c *cobra.Command, args []string) {
...
+// generate default HBA depending on DefaultSUReplAccessMode.
+func (p *PostgresKeeper) generateHBA(cd *cluster.ClusterData, db *cluster.DB) []string {
+
+	// Minimal entries for local normal and replication connections needed by the stolon keeper
+	// Matched local connections are for postgres database and suUsername user with md5 auth
+	// Matched local replication connections are for replUsername user with md5 auth
+	computedHBA := []string{
+		fmt.Sprintf("local postgres %s %s\n", p.pgSUUsername, p.pgSUAuthMethod),
+		fmt.Sprintf("local replication %s %s\n", p.pgReplUsername, p.pgReplAuthMethod),
+	}
```
sgotti: remove `\n`

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,86 @@ func keeper(c *cobra.Command, args []string) {
...
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "0.0.0.0/0", p.pgSUAuthMethod),
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "::0/0", p.pgSUAuthMethod),
+			fmt.Sprintf("host replication %s %s %s", p.pgReplUsername, "0.0.0.0/0", p.pgReplAuthMethod),
+			fmt.Sprintf("host replication %s %s %s", p.pgReplUsername, "::0/0", p.pgReplAuthMethod),
+		)
+
+	// Case strict access mode
+	case cluster.SUReplAccessStrict:
+		// allow for all servers su access
+		// we should not allow replication acces for 127.0.0.1 since it can be then forwarded through proxy.
```
sgotti: Sorry but I forgot why this is needed. The comment isn't really clear and doesn't looks related to the added rules.

cmd/keeper/keeper.go
t0k4rt: yes that's better !

cmd/keeper/keeper.go
t0k4rt: True,  it's changed

cmd/keeper/keeper.go
t0k4rt: Done

cmd/keeper/keeper.go
```diff
@@ -1754,3 +1757,83 @@ func keeper(c *cobra.Command, args []string) {
...
+			fmt.Sprintf("host replication %s %s %s", p.pgReplUsername, "0.0.0.0/0", p.pgReplAuthMethod),
+			fmt.Sprintf("host replication %s %s %s", p.pgReplUsername, "::0/0", p.pgReplAuthMethod),
+		)
+
+	// Case strict access mode
+	case cluster.SUReplAccessStrict:
+		computedHBA = append(
+			computedHBA,
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "127.0.0.0/8", p.pgSUAuthMethod),
+			fmt.Sprintf("host all %s %s %s", p.pgSUUsername, "::1/128", p.pgSUAuthMethod),
```
sgotti: Sorry but I don't remember why these line are needed? If really needed a comment explaining the reason will be useful.

sgotti:
@t0k4rt Can you import the unit test I made in https://github.com/sgotti/stolon/commit/cfc5a03709793f0a76ba5fa5aa95d339a6286330 and squash and rebase against the current master? Thanks!

cmd/keeper/keeper.go
t0k4rt: Actually It's not mandatory since we already have
`fmt.Sprintf("local postgres %s %s", p.pgSUUsername, p.pgSUAuthMethod)`
I can drop this

t0k4rt:
@sgotti Hello, I couldn't manage to cherry pick your commit so I included the test manually. I hope it's okay.

cmd/keeper/keeper.go
t0k4rt: Done

sgotti:
@t0k4rt unfortunately you have to rebase you branch on the current master and it'll be better if you squash all your commits in only one commit. If you have some problems doing it let me know.

t0k4rt:
@sgotti I think this should be ok, could you check ?

sgotti:
@t0k4rt can you please squash you commits in an unique one?

t0k4rt:
@sgotti I'm sorry, I merged my branch with your master to resolve conflict and squashed commits with rebase. But there are still conflict. Could you help me ?

sgotti:
@t0k4rt I rebased your branch using `git --rebase origin/master`, fixed two conflicts and did some minor changes to the doc, some comments and the commit title.

Take it here: https://github.com/sgotti/stolon/commit/1f4d4296394ab3b5863873c3ac3f388f4b8aa120

sgotti:
@t0k4rt sorry but it's not just my commit (I see 1157 changed files) so it's not rebased on current origin/master.

You should do something like this:

on your local branch:

`git fetch`
`git reset --hard origin/master`

then cherry pick my commit. The if you run `git rebase origin/master` you shouldn't see any error and change.

Then force push your branch to github to update the PR.

t0k4rt:
@sgotti is it okay now ? I started again from sorintlab/stolon master branch.

sgotti:
@t0k4rt We're almost there. 

I just noticed that the new functions in keeper.go are at the end of the file. I'll prefer to move them before but it's not a big deal.

I also see a difference in two comments:

```
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -187,9 +187,9 @@ type StandbySettings struct {
 type SUReplAccessMode string
 
 const (
-       // Allow access from every host
+       // Allow hba access to all members (0.0.0.0/0)
        SUReplAccessAll SUReplAccessMode = "all"
-       // Allow access from standby server IPs only
+       // Allow hba access to standby server only (strict mode)
        SUReplAccessStrict SUReplAccessMode = "strict"
 )
```

t0k4rt:
@sgotti I put all *PostgresKeeper functions with the others and put the right comments. This should look good !

sgotti:
@t0k4rt Great! Thanks a lot of work and patience! Merging.

t0k4rt:
@sgotti Grazie mille !

PR #414, opened 2018-01-21
build: allow explicit version configuration

abn:
When using source tarball, builds fail when discovering current version. This change allows for STOLON_VERSION, if set in the environment, to be used as the build version instead of attempting discovery.

The original issue was identified when building v0.8.0 RPMs. See [source](https://github.com/abn/stolon-rpm) for current workaround.

### DIFF
--- a/build
+++ b/build
@@ -43,7 +43,7 @@ mkdir -p ${BINDIR}
 
 export GO15VENDOREXPERIMENT=1
 
-VERSION=$(${BASEDIR}/scripts/git-version.sh)
+VERSION=${STOLON_VERSION:-$(${BASEDIR}/scripts/git-version.sh)}
 LD_FLAGS="-s -X ${REPO_PATH}/cmd.Version=$VERSION"
 
 # for static compilation we need to compile with cgo disabled (CGO_ENABLED=0),


sgotti:
@abn Thanks for the PR.

The idea, since I'm lazy and I wanted to avoid manually updating a version string every time, was to just use git info. Unfortunately this is breaking, as you noticed, build from a source tarball.

Your patch LGTM. Perhaps in a future we could just release ourself some source tarballs with the version already set? Do you have better ideas?

abn:
@sgotti I am all for laziness. I reckon that is a reasonable solution to provide hard coded versions, however will require additional finagling of the source before packing it up outside of SCM. Personally, I am not a big fan of that as it deviates from the tagged source and adds additional steps when preparing a release.

The other option is to use a tool like [bumpversion](https://pypi.python.org/pypi/bumpversion), so the tagging process inherently maintain a "<next-version>-devel" version for master and update that to the release version when tagging. Obviously trade off here is that every build from master will have the same version and won't include the hash.

The fix in the patch, just leaves that decision up to the packager. If they are using a git source, they can rely on the `git info` else, well enforce the version they want.

Hope that is helpful. :)

sgotti:
@abn Thanks for you answer and suggestions! Merging. Thanks!

abn:
Thanks for the merge @sgotti

nh2:
@abn Thanks a lot for this, this makes it much easier to package stolon for NixOS, where builds need to be reproducible (and the contents of `.git` pack files aren't guaranteed to be reproducible).

PR #416, opened 2018-01-22
Add option to disable color in logs

### LINKED ISSUE #415: Disable color in log output
### Submission type

Request for enhancement (RFE)

### Environment

Anything where log is not shown on a terminal

#### Stolon version

master

I'm using stolon inside a kubernetes with logs sent to Elasticsearch using fluentd (logs from container) and shown using Kibana

### Enhancement Description

Logging is forcefully using color in the output. This work well when show directly in a terminal, not with any other method (e.g. "| less" or in Kibana).
When using another method, the escape sequence are shown: "[34mINFO[0m	proxy/proxy.go:275	proxying to master address[...]"

I think this is important, as using Elasticsearch/Kibana is a common setup in Kubernetes and those escape sequence make the log less readable.
Even worse, when using ElasticSearch/Kibana, that means that the search string "error" won't match "[31mERROR[0m	sentinel/sentinel.go:274[...]" because by default is search for the *word* error (here the word is 31mERROR).

I could think of 3 ways to fix this:
* Do some guessing, if attached to a tty use color, if not don't use color (I don't like that solution, seems fragile)
* Always disable color
* Add an option to disable color in output

I could try to come with a PR, just need to know which solution should be done :)

PierreF:
This PR will fix #415 
It's not currently working (color are always disabled for now).

This issue is that I didn't find a way to pass option to pkg/log before logger is build. Logger initialization happens during init()/package import... while parsing option happens in main().
It does not seems to be dynamically updateable like is the level. To update the value, I need to build a new logger (which means that I would need to update all place where old logger is used).

I could think of the following way to solve this issue:
* After argument are parsed, create a logger with correct color settings and update place where logger is used (cmd/sentinel.log & pkg/postgresql.log for stolon-sentinel)
* Use go-isatty directly in init() of pkg/log and don't use flag to forcefully disable color but maybe environment variable

I'm not fan of either solutions that both looks hack. I'm open to better suggestion.


sgotti:
@PierreF Thanks for the PR! Yeah, since we cannot in place update the encoder I think we should define in `pkg/log` two kind of loggers (with and without color), by default set the global logger as a non colored logger and replace the it with a colored logger in the main. 

For the subpackages (now only `pkg/postgres`) we could add a global function (`SetLogger`) to set/update its global logger (like already done here: https://github.com/sorintlab/stolon/blob/master/cmd/proxy/proxy.go#L368) or just pass the logger in the constructor.

PierreF:
Updated PR to have two loggers. By default the non-colored one and if condition are meet, replace logger with the colored one.
Condition are:
* User provided the --log-level=true
* OR, user didn't provided --log-level=false and stderr is a tty

cmd/keeper/keeper.go
```diff
@@ -115,6 +117,7 @@ func init() {
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUUsername, "pg-su-username", user, "postgres superuser user name. Used for keeper managed instance access and pg_rewind based synchronization. It'll be created on db initialization. Defaults to the name of the effective user running stolon-keeper. Must be the same for all keepers.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPassword, "pg-su-password", "", "postgres superuser password. Only one of --pg-su-password or --pg-su-passwordfile must be provided. Must be the same for all keepers.")
 	cmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPasswordFile, "pg-su-passwordfile", "", "postgres superuser password file. Only one of --pg-su-password or --pg-su-passwordfile must be provided. Must be the same for all keepers)")
+	cmdKeeper.PersistentFlags().BoolVar(&cfg.logColor, "log-color", false, "enable color in log output (default if attached to a terminal)")
```
sgotti: Perhaps we could move this options and related logic inside the common command config: https://github.com/sorintlab/stolon/blob/master/cmd/common.go to avoid replicating it.

sgotti:
@PierreF LGTM! Let me know if you have other pending changes or I can merge it.

PierreF:
No, It's good for me.

sgotti:
@PierreF Thanks a lot! Merged.

PR #419, opened 2018-01-24
Docker image

marech:
Added Dockefile for 

- postgres:9.6
- postgres:10

Would be nice to integrate them in main repository so users can build image pretty fast.   
Also could think about automatic builds and push to  docker store so  users don't need to build image but simply pull it from docker store.


sgotti:
@marech We already have example docker images for the kubernetes example: https://github.com/sorintlab/stolon/tree/master/examples/kubernetes#docker-image which is very similar to you proposed images but templated for multiple versions. We also have images for the docker examples: https://github.com/sorintlab/stolon/tree/master/examples/docker#docker-images

We deliberately don't have a root and official Dockerfile because IMHO it doesn't makes sense, stolon can be deployed on multiple different enviroments, users have the need to add additional extensions, tools (like wal-g, wal-e etc...).

So there isn't an image that fits all the needs and for this reason providing an "official" image is a big burden since people will then ask to add extensionX, toolY etc and other will ask to remove them...

marech:
Ahh, i see, that's a good reason indeed :)

sgotti:
@marech Thanks!

PR #420, opened 2018-01-24
*: document how we use synchronous_standby_names syntax

sgotti:
And explain why we don't use postgres FIRST or ANY methods with N different than
the number of the wanted synchronous standbys but we have a different logic to
maintain consistency.


sgotti:
bad branch name...

PR #425, opened 2018-01-26
*: Make the store prefix configurable

sgotti:
Add a `--store-prefix` option to specify the store prefix.

In the meantime, for consistency, rename StoreBasePath to StorePrefix.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/sorintlab/stolon/common"
+
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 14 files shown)


kemko:
It seems that keeper has ignored --store-prefix key. I changed keeper-entrypoint.sh by add `--store-prefix=/stolon/cluster` but got `no cluster data available` error and in ps stolon-keeper miss this option.

sgotti:
@kemko I tested it and it works. Are you sure you built stolon with this PR included?

kemko:
@sgotti I forgot to add backslash while adding --store-prefix option and bash were interpreted it as another one command. Now it works.

sgotti:
@kemko Thanks! merged.

PR #433, opened 2018-02-07
*: add kubernetes store backend

sgotti:
This PR adds a new kubernetes store backend.

See the related doc in `/doc/architecture.md` and the updated kubernetes example
that will use this backend.

**TODO**
- [x] Add some CI tests: it'll take some time to update the test suite to be able run all the current tests inside a k8s cluster (it need to be changed to manage pod resources instead of simple processes, and also retrieve all the logs for debug purposes). So, as a first start, I'll only add tests to deploy a k8s stolon cluster and check that the cluster initializes correctly.


Flowkap:
Is this due for the next release?

sgotti:
@Flowkap We'd like to do a release (v0.10.0) next week. This feature is quite new so if you it'll be great if you can help us testing it. I'd like to improve the test suite to also cover all of the current integration tests on kubernetes but it'll take some more time.

Flowkap:
I'd tried to update the stolon chart,but failed due to not connecting sentinels. MAybe you know whats wrong there. 

https://github.com/lwolf/stolon-chart/issues/18#issuecomment-370433524

Flowkap:
Maybe one of you guys can take a look at the above mentione issue? I still can't find an obvious error in the generated config by that chart modification. Still all components do not kick into action as if they can't find the shared configmap. I did a side by side comparison of the chart generated stuff and your example. They look the same for me.

sgotti:
@Flowkap Do you have problems only with the chart or also with the stolon kubernetes example? If you also have the same problems with the stolon k8s example can you please open a new issue here on stolon since this PR is already merged?

Flowkap:
The example works fine. I tried to add the kubernetes backend options to the chart.ON the issue there I added thorough infos on what I did. I can't find a reason for the behaviour I documented there. I need to admit that Kubernetes and helm are all new to me so it might be a simple thing. :/

Flowkap:
I found it. Long story short: The app labels MUST be exactly "stolon-keeper" "stolon-proxy" and "stolon-sentinel" ...
sorintlab/stolon@38ae6b1#diff-95cdd374e9440fde010ff35d65f8cf3fR54

I'll add a proper PR on the chart later. You should really consider querying by a combination of component label and cluster-name, as a fully qualified app label should be best practise.

sgotti:
> You should really consider querying by a combination of component label and cluster-name, as a fully qualified app label should be best practise.

This is already done: https://github.com/sorintlab/stolon/blob/master/pkg/store/k8s.go#L67

If this isn't what you mean please open a new issue.

PR #455, opened 2018-03-09
Add PAT/NAT config

nabbar:
Hello,

First, thanks a lot for your work.

I needed to use a NAT / PAT configuration for postgres instances managed by the keeper. Also, I allowed myself to modify the code a bit to allow these features.

I named the flag as "advertise" which are optional flags. 
This advertise flag are added for listening pg address and listening pg port.

The flags are managed as : 
* If the listening is not set, it will use the advertise value. 
* If the advertise is not set, it will use the listening value. 
* if no one (advertise and listening) are not set, the keeper will return an error. 

I hope this code will be consistent with your idea of the project. 
I tried to make as few changes as possible but with the greatest flexibility possible.

Regards,

Nicolas.


nabbar:
Sorry, i forget to squash my commit...

sgotti:
@nabbar Thanks a lot of your PR! I will be able to deeply review it in the next week.

> Sorry, i forget to squash my commit...

Don't worry, you can squash it later after the review or when you prefer.

nabbar:
git squash done

cmd/keeper/cmd/keeper.go
```diff
@@ -288,7 +292,13 @@ func (p *PostgresKeeper) createPGParameters(db *cluster.DB) common.Parameters {
 		parameters[k] = v
 	}
 
-	parameters["listen_addresses"] = fmt.Sprintf("127.0.0.1,%s", p.pgListenAddress)
+	var lst = make([]string, 0)
+
+	if p.pgListenAddress != "*" {
+		lst = []string{"127.0.0.1"}
+	}
```
sgotti: just noticed that listening on localhost is probably not needed anymore by the stolon keeper since we use unix sockets. Perhaps it's useful for people doing local connections (local backups scripts etc...) so we could keep it to avoid possible regression.

cmd/keeper/cmd/keeper.go
```diff
@@ -1812,25 +1826,35 @@ func keeper(c *cobra.Command, args []string) {
 		die("cannot create data dir: %v", err)
 	}
 
-	if cfg.pgListenAddress == "" {
-		die("--pg-listen-address is required")
+	if cfg.pgAdvertiseAddress == "" && cfg.pgListenAddress == "" {
+		die("--pg-listen-address and/or --pg-advertise-address is required")
+	} else if cfg.pgListenAddress == "" {
+		cfg.pgListenAddress = "*"
```
sgotti: I'll prefer the user to always specifiy the listen address and not default to '*' when not specified.

cmd/keeper/cmd/keeper.go
```diff
@@ -1812,25 +1826,35 @@ func keeper(c *cobra.Command, args []string) {
...
 	}
 
-	ip := net.ParseIP(cfg.pgListenAddress)
+	if cfg.pgAdvertisePort == "" && cfg.pgPort == "" {
+		die("--pg-port and/or --pg-advertise-port is required")
+	} else if cfg.pgPort == "" {
+		cfg.pgPort = cfg.pgAdvertisePort
+	} else if cfg.pgAdvertisePort == "" {
+		cfg.pgAdvertisePort = cfg.pgPort
+	}
```
sgotti: This looks strange since if pgPort is not specified it'll be set to advertise port and if pgAdvertisePort is not specified it'll be set to pgPort. But pgPort has a default to 5432 so it'll probably never happen.

I'll prefer to keep the original behavior and just set pgAdvertisePort to pgPort when pgAdvertisePort is not specified.

cmd/keeper/cmd/keeper.go
```diff
@@ -1812,25 +1826,35 @@ func keeper(c *cobra.Command, args []string) {
...
+
+	ip := net.ParseIP(cfg.pgAdvertiseAddress)
 	if ip == nil {
-		log.Warnf("provided --pgListenAddress %q: is not an ip address but a hostname. This will be advertized to the other components and may have undefined behaviors if resolved differently by other hosts", cfg.pgListenAddress)
+		log.Warnf("provided --pg-advertise-address %q: is not an ip address but a hostname. This will be advertized to the other components and may have undefined behaviors if resolved differently by other hosts", cfg.pgAdvertiseAddress)
```
sgotti: The log can look strange to an user if it hasn't provided pgAdvertiseAddress. Perhaps to could be change to 
`PGAdvertiseAddress: %q: ...` and when setting pgAdvertiseAddress to pgListenAddress logging that we are setting PGAdvertiseAddress to PGListenAddress.

cmd/keeper/cmd/keeper.go
nabbar: yes i think so, but I wanted to keep as much functionality as possible.
I your are agree, I will remove it :)

cmd/keeper/cmd/keeper.go
nabbar: oki, no problem, I'm quizzing myself for a while about the default values to apply.

cmd/keeper/cmd/keeper.go
nabbar: ok, no problem :)

cmd/keeper/cmd/keeper.go
nabbar: Yes as following previous comment, I think this must be change to apply correct message.

nabbar:
I will do this fix, this WE or next week.
I keep you inform when i was ready.

sgotti:
@nabbar Is it ready? If you want I can help you to write the tests.

nabbar:
Hello,

Sorry, I am little busy this days... 
If you have time to update the test, I could be great :)

Thanks in advance

daMupfel:
Hi,

Is there any reason this was never finished? Is there anything i could do to push this?

Regards David

PR #456, opened 2018-03-11
Suggest Correctness with flag --store-key

nabbar:
Hello,

Other flag for file are --store-cert-file --store-ca-file, but for key, the word file is missing.
I suggest to add a new flag correctly write and keep old flag but with a deprecated mark.

- add flag store-key-file
- allow use store-key-file or store-key
- add deprecated flag store-key

I hope this code is compliant with your idea of the project and this difference who not do in some reason. 

Regards,

Nicolas.


nabbar:
Yes i understand your suggest :)
Great Idea. I will modify my fix this WE to do this :)

nabbar:
Done with cmd.PersistentFlags().Changed(...) method

nabbar:
rebase to last master

cmd/common.go
```diff
@@ -55,7 +55,15 @@ func AddCommonFlags(cmd *cobra.Command, cfg *CommonConfig) {
 	cmd.PersistentFlags().StringVar(&cfg.StoreEndpoints, "store-endpoints", "", "a comma-delimited list of store endpoints (use https scheme for tls communication) (defaults: http://127.0.0.1:2379 for etcd, http://127.0.0.1:8500 for consul)")
 	cmd.PersistentFlags().StringVar(&cfg.StorePrefix, "store-prefix", common.StorePrefix, "the store base prefix")
 	cmd.PersistentFlags().StringVar(&cfg.StoreCertFile, "store-cert-file", "", "certificate file for client identification to the store")
+
 	cmd.PersistentFlags().StringVar(&cfg.StoreKeyFile, "store-key", "", "private key file for client identification to the store")
+	cmd.PersistentFlags().MarkHidden("store-key")
+	cmd.PersistentFlags().MarkDeprecated("store-key", "use --store-key-file instead")
+
+	if !cmd.PersistentFlags().Changed("store-key") {
```
sgotti: Calling Changed() in the init method (since AddCommonFlags is called inside init() ) will always return false. My idea was to add both flags here (with store-key marked as deprecated) and in `CheckCommonConfig` verify (using the Changed() method) that not both options are changed or return an error.

nabbar: @sgotti ok, I will try to do it today.

nabbar:
@sgotti fix done. Please review and tell me if it's ok for you.

cmd/common.go
```diff
@@ -49,13 +49,20 @@ type CommonConfig struct {
...
 func AddCommonFlags(cmd *cobra.Command, cfg *CommonConfig) {
 	cmd.PersistentFlags().StringVar(&cfg.ClusterName, "cluster-name", "", "cluster name")
 	cmd.PersistentFlags().StringVar(&cfg.StoreBackend, "store-backend", "", "store backend type (etcdv2/etcd, etcdv3, consul or kubernetes)")
 	cmd.PersistentFlags().StringVar(&cfg.StoreEndpoints, "store-endpoints", "", "a comma-delimited list of store endpoints (use https scheme for tls communication) (defaults: http://127.0.0.1:2379 for etcd, http://127.0.0.1:8500 for consul)")
 	cmd.PersistentFlags().StringVar(&cfg.StorePrefix, "store-prefix", common.StorePrefix, "the store base prefix")
 	cmd.PersistentFlags().StringVar(&cfg.StoreCertFile, "store-cert-file", "", "certificate file for client identification to the store")
-	cmd.PersistentFlags().StringVar(&cfg.StoreKeyFile, "store-key", "", "private key file for client identification to the store")
+
+	cmd.PersistentFlags().StringVar(&deprecatedKeyFile, "store-key", "", "private key file for client identification to the store")
+	cmd.PersistentFlags().MarkHidden("store-key")
```
sgotti: MarkHidden isn't needed since MarkDeprecated already hides the flag.

cmd/common.go
```diff
@@ -74,7 +81,11 @@ func AddCommonFlags(cmd *cobra.Command, cfg *CommonConfig) {
 	}
 }
 
-func CheckCommonConfig(cfg *CommonConfig) error {
+func CheckCommonConfig(cmd *cobra.Command, cfg *CommonConfig) error {
+	if cmd.PersistentFlags().Changed("store-key") {
```
sgotti: I'll also add a check before this line that will return an error if both are provided (to avoid any ambiguity):

```
	if cmd.PersistentFlags().Changed("store-key") && cmd.PersistentFlags().Changed("store-key-file") {
		return fmt.Errorf("only one of --store-key and --store-key-file should be provided")
	}
```

PR #458, opened 2018-03-11
Add Socket Dir Management 

nabbar:
Hello,

In some case, the /tmp is mount as noexec. The socket can have some trouble to work correctly.
So, I have add a little addition to offert the socket location capability.

Indeed, I have extend the auth-method to having different auth method between socket connection and remote connection. 

Add Socket Dir Management :
- add an option flags to specify the socket dir. Default value is the common.PgUnixSocketDirectories. If not specify or cannot access, use the default value;
- add option for defined the socket authmethod for SU/Repl to allow differt access in socket and in remote;

Regards,

Nicolas.


sgotti:
@nabbar Thanks for the PR! I'll prefer to split the PR in two. One for setting the socket dir (this one) and a new PR for setting different auth methods for the socket since it takes more time for review the changes and will need some tests.

nabbar:
> I'll prefer to split the PR in two. One for setting the socket dir (this one) and a new PR for setting different auth methods for the socket since it takes more time for review the changes and will need some tests.

Sure, I will prepare this this WE I think.

nabbar:
@sgotti apply rebase from master

volfco:
@nabbar any update on getting this merged?

sgotti:
@volfco This PR is almost two years old. Feel free to create a new PR to implement this feature is you need it.

PR #459, opened 2018-03-14
doc: add postgresl SSL setup

sgotti:
/cc @nabbar


nabbar:
great :) 
I will see that this WE or next week.

PR #462, opened 2018-03-23
Flags to specify auth method for local connection

nabbar:
Hello 

As suggest, I have split the PR #458 with this.
This PR is so just to the new flag to specify a different auth method between remote and local.

Extend Socket Management :
- add option to specify the socket (local) auth method for SU/Repl 
- use this auth method for the local connection params function
- use value of auth method flag if local auth method flag is not specify

Thanks,

Nicolas JUHEL.


nabbar:
@sgotti apply rebase from master

PR #475, opened 2018-04-16
sentinel: update all the various changetimes.

sgotti:
Some clusterdata change time weren't set or updated, improve the sentinel to always set/update
them when created/changed.

Also add a global clusterdata changetime updated when the clusterdata is
written.


pkg/cluster/cluster.go
```diff
@@ -690,11 +690,12 @@ type ClusterData struct {
...
-	Cluster       *Cluster
-	Keepers       Keepers
-	DBs           DBs
-	Proxy         *Proxy
+	FormatVersion uint64    `json:"formatVersion"`
+	ChangeTime    time.Time `json:"changeTime"`
+	Cluster       *Cluster  `json:"cluster"`
+	Keepers       Keepers   `json:"keepers"`
+	DBs           DBs       `json:"dbs"`
+	Proxy         *Proxy    `json:"proxy"`
```
nh2, line 698: This introduced a backwards-incompatible change that was not announced in the changelog.

It broke our monitoring scripts that relied on the `stolonctl clusterdata` JSON key `.Keepers`, which is now `.Keepers`.

It may make sense adding that information to the next release notes retrospectively.

PR #477, opened 2018-04-17
keeper: use wal streaming during pg_basebackup

sgotti:
To avoid replica creation failing due to removed wal on the source.

Also double the number of wal senders since wal streaming will use an additional
wal sender.

On postgres >= 9.6 we'll provide a replication slot to use.

Note that on postgres 10 we could avoid passing a replication slot since it'll
create a temporary one. For the moment use the persistent replication slot that
will be created by the keeper on the instance also on postgres 10. Sometimes the
initial pg_basebackup will fail if the replication slot is not yet created but
the successive try will succeed.


sgotti:
@alesz Can you try it to see if it works for your case?

alesz:
Tested, works without any issues, cheers

sgotti:
@alesz Thanks! Merging.

PR #482, opened 2018-05-01
swarm example

Nowheresly:
Hi @sgotti 
I was looking for a solution for postgres HA inside docker. So I decided to evaluate your project and played a little bit with your very well documented kubernetes example.
At the end, I was able to create a simple swarm version based on the kubernetes example.
So in case you could be interested in reviewing this swarm example, I would really appreciate to get your feedback.


sgotti:
@Nowheresly Thanks for the PR!

We already have docker swarm example here. https://github.com/sorintlab/stolon/tree/master/examples/docker contributed by @ihcsim . I'm not sure how to proceed, is your example very different from the current one?

Nowheresly:
Thanks for the reply.
Yes, I already looked at it. But it seemed to me not as up-to-date as the kubernetes example IMO.
This PR uses docker secret, rolling upgrades and network overlay as far as swarm is concerned while using the same docker image as the one in use with the kubernetes example.
If @ihcsim could have a look at this PR, I would be glad to read his comments as well.

ihcsim:
@Nowheresly You can just remove my outdated example in your PR.

Nowheresly:
Based on the comment of @ihcsim, I just added a commit to remove the previous swarm example.
i can squash the two commits if needed.

build
```diff
@@ -98,7 +98,7 @@ cp ${GOPATH}/bin/keeper ${BINDIR}/stolon-keeper
 
 # Copy binaries to Dockerfile image directories
 declare -a DOCKERFILE_PATHS
-DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker ${BASEDIR}/examples/docker)
+DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker ${BASEDIR}/examples/swarm/image/docker ${BASEDIR}/examples/docker)
```
ihcsim: `${BASEDIR}/examples/docker` is removed, right?

examples/swarm/image/docker/Dockerfile.template
```diff
@@ -0,0 +1,9 @@
+FROM postgres:${PGVERSION}
```
ihcsim: Isn't this template the same as the one in `examples/kubernetes/image`?

examples/swarm/image/docker/Makefile
```diff
@@ -0,0 +1,20 @@
+ifndef PGVERSION
```
ihcsim: Isn't this Makefile the same as the one in `examples/kubernetes/image`?

examples/swarm/image/docker/Dockerfile.template
Nowheresly: Yes indeed. The content of both `examples/kubernetes/image` and `examples/swarm/image` are the same. My goal with this PR was to update the swarm example base on the kubernetes example. So I chose to use the same docker images for both examples.

examples/swarm/image/docker/Makefile
Nowheresly: same reply as above

Nowheresly:
Thanks @ihcsim for your review!

ihcsim:
@sgotti This PR LGTM.

examples/swarm/image/docker/Dockerfile.template
sgotti: @Nowheresly I'd like to avoid maintaining multiple set of identical files. I think we could remove these files and update the documentation to point to the examples/kubernetes/image.

.gitignore
```diff
@@ -14,3 +14,5 @@ Session.vim
 bin/
 /gopath/
 /release/
+
+.idea/
```
sgotti: Don't add changes to .gitignore in this PR.

build
```diff
@@ -98,7 +98,7 @@ cp ${GOPATH}/bin/keeper ${BINDIR}/stolon-keeper
 
 # Copy binaries to Dockerfile image directories
 declare -a DOCKERFILE_PATHS
-DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker ${BASEDIR}/examples/docker)
+DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker ${BASEDIR}/examples/swarm/image/docker
```
sgotti: Since I'd like to keep only one image let's remove this.

examples/swarm/docker-compose-pg.yml
```diff
@@ -0,0 +1,86 @@
...
+      replicas: 2
+      update_config:
+        parallelism: 1
+        delay: 30s
+        order: stop-first
+        failure_action: rollback
+
+volumes:
+  pgkeeper1:
+    driver: local
```
sgotti: Does using the local volume plugin will create some problems since on multiple node cluster since the keeper cannot be started on another node than the one where the volume has been created? Perhaps some notes to change this to a real persistent volume will be good.

.gitignore
Nowheresly: Sure, this can be done in another PR

build
Nowheresly: Ok, done

examples/swarm/docker-compose-pg.yml
Nowheresly: You're right, this section deserves probably more explanations. 
Out of the box, only the local driver is available with docker swarm. That's the reason why this swarm example is based on the local driver. 
I specifically chose to provide different names for volumes of keeper1 and keeper2 so that we can't have any volume mismatch if keeper1 and keeper2 are deployed on the same node of the swarm.
Nevertheless, IMHO, it makes no sense to deploy all keepers on the same node. So I would recommend to set some constraints on both keepers so that each node would have its own keeper. Here is an example of of a configuration where keeper1 will be deployed only on a node whose nodename label is equal to node1:
```
  keeper1:
    deploy:
      placement:
        constraints: [node.labels.nodename == node1]
     ...
```
In case each keeper is deployed on its own node, a node failure should not prevent the cluster to remain in function.

I'm not sure if my explanations are clear to you. I would propose to add a comment in the README to explain this approach. @sgotti What do you think?

Of course, if a specific docker volume plugin is available that could enable to share volumes between nodes, the volume section of the docker-compose-pg.yml would have to be adapted accordingly to leverage this plugin and there's no need to set any constraints.

sgotti: @Nowheresly Your explanation is right and I agree with you. A comment in the README will be great!

Nowheresly: Thanks for the review

Nowheresly: @sgotti I updated the README to explain how to set labels on each node and how to set constraints on each keeper. Hopefully, it should make sense to swarm users.

build
```diff
@@ -98,7 +98,7 @@ cp ${GOPATH}/bin/keeper ${BINDIR}/stolon-keeper
 
 # Copy binaries to Dockerfile image directories
 declare -a DOCKERFILE_PATHS
-DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker ${BASEDIR}/examples/docker)
+DOCKERFILE_PATHS=(${BASEDIR}/examples/kubernetes/image/docker
```
sgotti: There's a missing closing bracket that is breaking the build

Nowheresly: Indeed, sorry for that!

PR #489, opened 2018-05-19
Add possibility to expose first slave via proxy

epustobaev:


sgotti:
@epustobaev Thanks for the PR! What's your use case to access just one random standby? Some people are doing this generating an haproxy conf to balance to all standbys from the cluster data since the proxy main role is to force connections to elected masters. Perhaps a little tool (i.e. a stolonctl command) to generate something similar and keep it updated should be better?

epustobaev:
Thanks for comment .

It was the simplest way to reproduce concrete project configuration with one read-only instance for some cases e.g. admin interfaces. And in the same time we're not using any balancing/proxy tools for database.

So we're using it already. And It just works and solves the problem.

But anyway I get your point and it sounds logical. If it not fits your architecture - I'll try to use your advice.

lawrencejones:
Yeah just another voice in this, being able to expose a replica, perhaps just the sync, is something we'd really want to do via stolon-proxy. We want the fencing to break connections whenever the sync moves, and we don't want to insert another tool (like haproxy) in the middle of our connections when we already have a combo of stolon and pgbouncer.

sgotti:
@epustobaev @lawrencejones

Balancing to slaves, has we noticed by the comments, can be implemented in many different ways and every user has different requirements. So at first the interested parties should defines which features are needed and then I'll suggest to implement it in another component outside the stolon-proxy (stolon-standbys-proxy). It could always be merged inside the stolon proxy when it'll stable.

sgotti:
@epustobaev @czernitko  @lawrencejones

Balancing to standbys, has we noticed by the comments, can be implemented in many different ways and every user has different requirements. So at first the interested parties should define which features are needed and then I'll suggest to implement it in another component outside the stolon-proxy (stolon-standbys-proxy). It could always be merged inside the stolon proxy when it'll be stable.

I don't see having different processes an issue, instead I'll prefer having different processes managing primary vs standby connections and in 2020 it's not so difficult to define a new service in systemd, k8s etc...

lawrencejones:
This is what we're running with atm: https://github.com/gocardless/stolon/pull/33

We run two stolon proxies per postgres node, one on port 7432 for the connection to the primary, another on 8432 for synchronous replica. It's working well for us.

PR #494, opened 2018-05-28
sentinel: fix check for same reported vs spec sync standbys

sgotti:
When electing a new master we are swapping the new master uid with the old
master uid in the syncstandbys slice. This could end with an unordered slice in
the spec that will make the sentinel fail the check that the reported standbys
are the same of the spec one blocking any future syncstandby update.

Since the reported order is not a problem just check that the syncstandbys are
the same regardless of their order.

We'll keep the sorting to avoid unneeded updates to synchronous_standby_names by
the keeper.

### DIFF
--- a/pkg/util/slice_test.go
+++ b/pkg/util/slice_test.go
@@ -1,0 +1,69 @@
+// Copyright 2018 Sorint.lab
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import "testing"
+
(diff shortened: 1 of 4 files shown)


pkg/util/slice_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}{
+		{[]string{}, []string{}, true},
+		{[]string{"", ""}, []string{""}, false},
+		{[]string{"", ""}, []string{"", ""}, true},
+		{[]string{"a", "b"}, []string{"a", "b"}, true},
+		{[]string{"a", "b"}, []string{"b", "a"}, true},
+		{[]string{"a", "b", "c"}, []string{"a", "b", "c"}, true},
+		{[]string{"a", "b", "c"}, []string{"b", "c", "a"}, true},
+		{[]string{"a", "b", "c", "a"}, []string{"a", "c", "b", "b"}, false},
+		{[]string{"a", "b", "c", "a"}, []string{"a", "c", "b", "b"}, false},
```
nh2: May want to add some cases where the two slices have different length to get coverage of the `if len(a) != len(b) { return false }` bit (which is what makes it correct e.g. for `{abc, abcd, false}`).

nh2:
That's great, thanks!

pkg/util/slice_test.go
sgotti: It's covered by the second test (the one with empty strings) but I'll add a more explicit test with non empty strings to make it clear.

PR #495, opened 2018-05-28
keeper: fix atomix writes.

sgotti:
The current atomic write functions inside pkg/postgresql are missing some error
checks and don't close the file before renaming. Improve common.WriteFileAtomic
to be able to stream writes and use it.

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -1678,7 +1678,7 @@ func (p *PostgresKeeper) saveKeeperLocalState() error {
 	if err != nil {
 		return err
 	}
-	return common.WriteFileAtomic(p.keeperLocalStateFilePath(), sj, 0600)
+	return common.WriteFileAtomic(p.keeperLocalStateFilePath(), 0600, sj)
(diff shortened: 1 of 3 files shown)


nh2:
I see now what you mean with the missing error checks in `f.WriteString()`:

`f.WriteString(fmt.Sprintf("%s = '%s'\n", n, v))` in `WriteRecoveryConf` should have `_, err = ` left of it.

I'm a bit annoyed at myself though, because I stared at that line for a long time and still didn't spot it.

I guess go makes it very easy to forget to check errors for functions that don't return values you need.

Change looks good to me!

nh2:
@sgotti What do you think about using a linter like https://github.com/kisielk/errcheck that promises to find all unused errors returned by functions?

sgotti:
@nh2 Thanks for the review!

> @sgotti What do you think about using a linter like https://github.com/kisielk/errcheck that promises to find all unused errors returned by functions?
Sure, will try to add it in another PR. Just noticed that my VIM setup with ALE + gometalinter isn't working since it should have reported it...

PR #500, opened 2018-06-01
Removing carriage return from example secret

whelgeson:
The example secret is encoded with a carriage return which causes problems with cluster auth when using the default example scripts. Although this password should be changed, users who want to quickly test out stolon in minikube will find that they are unable to authenticate into the cluster using password1.

### DIFF
--- a/build
+++ b/build
@@ -33,6 +33,9 @@ fi
 ORG_PATH="github.com/sorintlab"
 REPO_PATH="${ORG_PATH}/stolon"
 
+# Hack to be sure that:
+# * all the dependencies are vendored
+# * if cloned as another repo name it will compile anyway
 export GOPATH=${PWD}/gopath
 
 rm -f $GOPATH/src/${REPO_PATH}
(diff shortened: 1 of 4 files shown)


sgotti:
@whelgeson Thanks for the PR. It LGTM. k8s tests are failing because you based this on v0.10.0 tag. Can you please rebase on master?


Just curious: also with the carriage return it always worked correctly with the default k8s example since  in https://github.com/sorintlab/stolon/blob/master/cmd/keeper/cmd/keeper.go#L152-L168

`readPasswordFromFile` uses `strings.TrimSpace` that will remove the leading `\n`.

Where did you discovered the issue?

whelgeson:
This was with Minikube v0.25.0 on Windows 10 w/HyperV. I don't have the setup available right now as it's been a few months since I ran into it. But, what I did was expose stolon-proxy-service as a nodeport and tried to connect to the proxy service but found myself unable to authenticate into postgresql unless I pasted in the default password with a newline character.

sgotti:
@whelgeson Thanks for the explanation. Can you please squash the commits in an unique one based on master?

PR #505, opened 2018-06-07
Removing carriage return from example secret

whelgeson:
The example secret is encoded with a carriage return which causes problems with cluster auth when using the default example scripts. Although this password should be changed, users who want to quickly test out stolon in minikube will find that they are unable to authenticate into the cluster using password1.

Referring to #500


sgotti:
@whelgeson Thanks for the PR! looks like it's based on v0.10.0 tag yet and not on current master but I merged it anyway since it's a oneliner and I tested it locally.

PR #555, opened 2018-08-27
Use readlinkdashf function in test for cross compatibility with osx

jskswamy:
readlink -f option is not available in osx,
instead use greadlink package which is GNU readlink for osx if available
Homebrew provide a coreutils package containing greadlink (GNU readlink)

### DIFF
--- a/test
+++ b/test
@@ -11,7 +11,28 @@
 #
 set -e
 
-BASEDIR=$(readlink -f $(dirname $0))
+# Cross compatibility with osx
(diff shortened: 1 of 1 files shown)


sgotti:
@jskswamy This is already handled in `build` but was "lost" in `test`. I don't use Max OSX but I think that solution (also used in k8s) is better and doesn't require additional packages. Can you please just copy the `readlinkdashf` function also in test and use it?

sgotti:
@jskswamy Thanks a lot! LGTM! Merging.

PR #561, opened 2018-09-07
[WIP] Add command to restart all underlying pg

### LINKED ISSUE #88: stolonctl: implement postgres reload/restart

prabhu43:
Fixes #88 

This PR is to restart the underlying postgres instances of all keepers

## Reason

Few parameters like `max_connections` and `shared_buffers` require restart of underlying postgres in each Keeper. We always had to run a `pg_ctl stop -m fast -W` on the node/pod to restart the postgres (keeper only does a reload)

## Approach

*Keeper*

- Stolon Keepers watch for a flag in ClusterData.Keeper.Status if `schedulePgRestart` is set to true
- If true, a keeper would set it false and restart the underlying postgres using `pgManager`

*CLI*

- `stolonctl` has a subcommand called `pgrestart`
- It would set `schedulePgRestart` to `true` for all keepers (which would trigger a restart of all clusters)

## Objective

Initially, we wanted to achieve a restart with 0 downtime by promoting an existing slave as a master as present in #88 However, it takes 7 seconds to elect a new master. Simply triggering a fast restart of postgres only cause 1 second downtime.

Please share your comments and feedbacks.

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -712,6 +712,7 @@ func (p *PostgresKeeper) Start(ctx context.Context) {
 	endSMCh := make(chan struct{})
 	endPgStatecheckerCh := make(chan struct{})
 	endUpdateKeeperInfo := make(chan struct{})
+	endRestartPGCh := make(chan struct{})
 
 	var err error
 	var cd *cluster.ClusterData
@@ -739,6 +740,7 @@ func (p *PostgresKeeper) Start(ctx context.Context) {
 	smTimerCh := time.NewTimer(0).C
 	updatePGStateTimerCh := time.NewTimer(0).C
 	updateKeeperInfoTimerCh := time.NewTimer(0).C
+	restartPGTimerCh := time.NewTimer(0).C
 	for true {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


sgotti:
@prabhu43 Thanks for the PR, do you need a review or should I wait untile it's not a WIP and tests passes?

prabhu43:
@sgotti Please review the implementation and give your feedback. We marked it as WIP to write few more tests. We will fix the failing tests and write few more tests

cmd/keeper/cmd/keeper.go
```diff
@@ -779,10 +781,53 @@ func (p *PostgresKeeper) Start(ctx context.Context) {
 
 		case <-endUpdateKeeperInfo:
 			updateKeeperInfoTimerCh = time.NewTimer(p.sleepInterval).C
+		case <-restartPGTimerCh:
+			go func() {
+				p.restartPostgresIfScheduled(ctx)
```
sgotti: I will prefer to not add a new time and a new function the read clusterdata again, this will also cause some race condition when an user asks to restart postgres while the main loop is acting on it. So I'll do this in the main check.

dineshba:
@prabhu43 can we close this

PR #567, opened 2018-09-19
proxy: use context inside proxy check

sgotti:
Use context instead of a timer to handle checks and timeouts.


sgotti:
Closing, will create a new PR.

PR #568, opened 2018-09-21
Restart postgres if required when updating pgParameters

### LINKED ISSUE #88: stolonctl: implement postgres reload/restart

prabhu43:
Alternate approach that. Fixes #88 

## Context:

If some postgres configurations (pgParameters) that require restart are updated, a restart is triggered

## Approach:

For Postgres versions >= 9.5,

1. After reloading pgParameters, we check if `pg_settings` table has `pending_restarts`

> pending_restart | boolean | true if the value has been changed in the configuration file but needs a restart; or false otherwise.

[9.5 pg_settings reference](https://www.postgresql.org/docs/9.5/static/view-pg-settings.html)

For older versions of postgres,

1. Figure out the changed Parameters.
2. Check if any of the changedParameters requires restart by checking in `pg_settings`'s `context`

> postmaster: These settings can only be applied when the server starts, so any change requires restarting the server. Values for these settings are typically stored in the postgresql.conf file, or passed on the command line when starting the server. Of course, settings with any of the lower context types can also be set at server start time.

[9.4 pg_settings reference](https://www.postgresql.org/docs/9.4/static/view-pg-settings.html)


cmd/keeper/cmd/keeper.go
```diff
@@ -1588,6 +1589,17 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 		if err := pgm.Reload(); err != nil {
 			log.Errorw("failed to reload postgres instance", err)
 		}
+		needsRestart, err := pgm.IsRestartRequired(changedParams)
```
sgotti: Just a not for a future improvement (will prefer doing this in a separate PR): we could report the need restart in the PostgresState reported to the sentinel so it could be showed in the `stolonctl status` output for people that won't enable automatic restart (see below comment).

cmd/keeper/cmd/keeper.go
```diff
@@ -1588,6 +1589,17 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
...
+		if err != nil {
+			log.Errorw("Failed to checked if restart is required", err)
+		}
+
+		if needsRestart {
```
sgotti: I'll make this opt in adding an option in the cluster spec. I'll leave it disabled by default (at least for the moment, to not break current expected behavior and avoid possible issue when users provides a wrong parameter that will break instance restart)

sgotti:
@prabhu43 gentle ping.

cmd/keeper/cmd/keeper.go
prabhu43: @sgotti As you suggested, We will make the pgRestart optional(disbaled by default) and add some integration tests. We will try to close this by this weekend.
cc @aswinkarthik

sgotti:
> We added Integration test.

Great!

> However, we are not able to debug why it is failing exactly. The logs are too verbose. Could you give us some pointers?

You should fetch the output log from semaphore and search for `--- FAIL` : 

```
--- FAIL: TestAutomaticPgRestart (28.41s)
[...]
    config_test.go:550: expected max_connections 150 is not equal to actual 100

```

> Also we found it difficult to run it locally. Are there any docs for doing so?

We just do the same that is done on semaphore:

For example to test with the locally installed postgres using etcdv3 backend:

```
INTEGRATION=1 STOLON_TEST_STORE_BACKEND=etcdv3 bash -x ./test
```

You can also parallelize (if you have fast disks):

```
INTEGRATION=1 STOLON_TEST_STORE_BACKEND=etcdv3 PARALLEL=10 bash -x ./test
```

Or use a tmpfs if you have a lot of ram as I do:

```
mkdir /dev/shm/stolon
export TMPDIR=/dev/shm/stolon

rm -rf /dev/shm/stolon/* ; INTEGRATION=1 STOLON_TEST_STORE_BACKEND=etcdv3 PARALLEL=20 bash -x ./test
```
so the tests will complete in ~2 minutes.

aswinkarthik:
@sgotti The integartion test is done. Could you review the commit?

sgotti:
@aswinkarthik Great! Will review tomorrow.

sgotti:
@prabhu43 Looks good to me!

Just noticed a behavior that I'm not sure how to deal with:
* If you have `"automaticPgRestart": false` and you update a parameter that needs a restart, postgres won't be restarted (correct). If you then set `"automaticPgRestart": true` it won't be restarted since pgm.curParameters has been updated and isn't changed. You have to update the parameter again (with a different value) to trigger the automatic restart.

I don't see this as a big deal and the behavior could be improved in a next PR if needed. What do you think?

We should also update the related doc:

* https://github.com/sorintlab/stolon/blob/master/doc/cluster_spec.md: add the `automaticPgRestart` option.
* https://github.com/sorintlab/stolon/blob/master/doc/postgres_parameters.md: describe the new behavior.

We can also add this in a separate PR. Let me know what you prefer.

prabhu43:
@sgotti We felt that the intent of "automaticPgRestart" is to do postgres restart only when it is required which is on changing some pgParameters. 

We will add update the docs in this PR itself.

aswinkarthik:
I have updated documentation @sgotti

aswinkarthik:
> If you have "automaticPgRestart": false and you update a parameter that needs a restart, postgres won't be restarted (correct). If you then set "automaticPgRestart": true it won't be restarted since pgm.curParameters has been updated and isn't changed. You have to update the parameter again (with a different value) to trigger the automatic restart.

Regarding this, there is a slight inconsistency here based on Postgres version.

9.5 and after:

If `automaticPgRestart` is false, and `max_connections` is changed, then `automaticPgRestart` is enabled, the postgres **will be restarted**. As in the database, `pending_restart` is a column in `pg_settings` and it will be marked true and hence triggering a restart.

For 9.4 and lower:

If `automaticPgRestart` is false, and `max_connections` is changed, then `automaticPgRestart` is enabled, the postgres **will not be** **restarted**. As in the database, `pending_restart` is a column in `pg_settings` and it will be marked true and hence triggering a restart.

prabhu43:
@sgotti We updated the docs with required details. Can you review those changes and merge the PR ?

sgotti:
@prabhu43 @aswinkarthik LGTM. Merging. Thanks a lot!

prabhu43:
@sgotti Thanks for merging. When can we expect this to be released?

PR #579, opened 2018-10-27
Export cluster data into file

jskswamy:
Add support for `clusterdata read` **subcommand** to write clusterdata into a file

Usage example:

```sh
stolonctl clusterdata read --cluster-name stolon --store-backend consul --file cluster_data.json
```


dineshba:
@sgotti Updated the branch with respect to current master and squashed the commits. Please refer it

PR #591, opened 2018-12-10
Make wal_keep_segments configurable

vitalif:
Make wal_keep_segments configurable, at least 8 as it was previously

I need it to setup an additional non-stolon slave...

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -206,11 +206,25 @@ func (p *PostgresKeeper) walLevel(db *cluster.DB) string {
 	return walLevel
 }
 
+func (p *PostgresKeeper) walKeepSegments(db *cluster.DB) int {
+	walKeep := 8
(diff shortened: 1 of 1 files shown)


cmd/keeper/cmd/keeper.go
```diff
@@ -206,11 +206,25 @@ func (p *PostgresKeeper) walLevel(db *cluster.DB) string {
...
+	}
+	return walKeep
+}
+
 func (p *PostgresKeeper) mandatoryPGParameters(db *cluster.DB) common.Parameters {
 	return common.Parameters{
 		"unix_socket_directories": common.PgUnixSocketDirectories,
 		"wal_level":               p.walLevel(db),
-		"wal_keep_segments":       "8",
+		"wal_keep_segments":       fmt.Sprintf("%v", p.walKeepSegments(db)),
```
sgotti: use %d instead of %v

sgotti:
Superseded by #664

Issue #358, opened 2017-09-22
Feature request: JSON output from "stolonctl status"


charles-dyfis-net:
At present, parsing the output from "stolonctl status" requires ad-hoc text processing.

An option to provide JSON output would permit easier parsing (with tools such as `jq`), and provide a means for future extension (by adding additional keys without changing the meaning of existing ones) that wouldn't have potential to impact existing parsers.

With upstream approval/consensus, I'm willing to do the implementation myself.

sgotti:
@charles-dyfis-net `stolonctl status` just parses the clusterdata, so another option could be to just parse the `stolonctl clusterdata` json output that contains all the available information. This is what we usually do in our monitoring scripts or for debugging.

I'm not against providing a json output to `stolonctl status` but it I'll just be a sort of redundant and reworked/cleaned cluster data and will need to be versioned since it could change (the current status output could be improved but it's currently enough for our needs)

charles-dyfis-net:
Using `clusterdata` makes sense to me.

PR #628, opened 2019-03-27
Support JSON output in stolonctl status

hmac:
This change allows you to pass a `--format` flag to `stolonctl status` to specify the output format. The available options are `text` (the current output format) and `json`, which produces the same data but in a structured format. The idea is that you can build automation around this information without relying on the semi-structured human-readable format.


sgotti:
@hmac Thanks for your PR! I'll review it in the next days. 

As a note there was a related issue here: #358

hmac:
Thanks! And apologies, missed that issue!

hmac:
@sgotti thanks for the review and bug spot! I've updated this and it looks like CI has passed now.

sgotti:
@hmac Thanks! Merging.

PR #636, opened 2019-04-26
Show keeper sync type in status

dyson:
Adds the type of sync a keeper is to the stolonctl status output which can be really useful.

Currently we are doing this through passing the cluster data through jq which isn't ideal.


dyson:
Thanks @sgotti, thanks for looking at this! I've updated the PR to also support adding the role to the JSON formatted output.

cmd/stolonctl/cmd/status.go
```diff
@@ -315,19 +338,50 @@ func generateStatus() (Status, error) {
...
+			keeperRoles[dbuid] = "master"
+		} else if isSynchronousStandby(parentDbuid, dbuid, cd) {
+			keeperRoles[dbuid] = "sync"
+		} else if isExternalSynchronousStandby(parentDbuid, dbuid, cd) {
+			keeperRoles[dbuid] = "external sync"
+		} else {
+			keeperRoles[dbuid] = "async"
+		}
+		db := cd.DBs[dbuid]
+		followers := db.Spec.Followers
```
sgotti: Instead of making a recursive function can you just iterate over all dbs and use their `followConfig` to detect the parent?

cmd/stolonctl/cmd/status.go
```diff
@@ -315,19 +338,50 @@ func generateStatus() (Status, error) {
...
+	buildRoles = func(parentDbuid, dbuid string, cd *cluster.ClusterData, keeperRoles map[string]string) map[string]string {
+		if dbuid == cd.Cluster.Status.Master {
+			keeperRoles[dbuid] = "master"
+		} else if isSynchronousStandby(parentDbuid, dbuid, cd) {
+			keeperRoles[dbuid] = "sync"
+		} else if isExternalSynchronousStandby(parentDbuid, dbuid, cd) {
+			keeperRoles[dbuid] = "external sync"
+		} else {
+			keeperRoles[dbuid] = "async"
+		}
```
sgotti: There can be also a case where a keeper is not set to be a standby (when exceeding MaxStandbys) si it should be noted as "none"

cmd/stolonctl/cmd/status.go
```diff
@@ -282,12 +284,28 @@ func generateStatus() (Status, error) {
...
+	status.Cluster = cluster
+
 	keepers := make([]KeeperStatus, 0)
 	kssKeys := cd.Keepers.SortedKeys()
+	keeperRoles := getKeeperRoles(status.Cluster.MasterDBUID, cd)
 	for _, kuid := range kssKeys {
 		k := cd.Keepers[kuid]
 		db := cd.FindDB(k)
 		dbListenAddress := "(no db assigned)"
+		role := ""
```
sgotti: Better to set it to something like `none` instead of an empty value

dyson:
@sgotti no problem at all! 

Thanks for the review. We'll work through these and get it in shape. Cheers!

cmd/stolonctl/cmd/status.go
mos3abof: @sgotti, would setting the none value to `no role assigned` be acceptable? The reason of proposing this is that I found it a bit confusing displaying (none) next to the keeper. This is how it looks like in both cases:

With `none`:
```
===== Keepers/DB tree =====

keeper0 (master)
  keeper1 (sync)
  keeper2 (none)
```

With `no role assigned`:
```
===== Keepers/DB tree =====

keeper0 (master)
  keeper1 (sync)
  keeper2 (none)
```

Which do you think works better?

PR #637, opened 2019-04-29
Fix failover process after disk crash

maksm90:
I have encountered the problem on test mini installation (cluster based on raspberry pi) that autofailover didn't performed after disk (SD card, in case of raspberry) manual insulation.

The problem lies in `PostgresKeeper.updatePGState` routine (cmd/keeper/cmd/keeper.go:561) that Postgres state isn't updated in DCS (consul or etcd) when `PostgresKeeper.GetPGState` routine returns some not null error. Such error is emitted after check whether PGDATA is initialized (`Manager.IsInitialized` routine, cmd/keeper/cmd/keeper.go:641) what is induced after disk file system operation failure (my case).

The current pull request fixes this issue.

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -558,7 +558,6 @@ func (p *PostgresKeeper) updatePGState(pctx context.Context) {
 	pgState, err := p.GetPGState(pctx)
 	if err != nil {
 		log.Errorw("failed to get pg state", zap.Error(err))
-		return
 	}
 	p.lastPGState = pgState
 }
@@ -641,7 +640,7 @@ func (p *PostgresKeeper) GetPGState(pctx context.Context) (*cluster.PostgresStat
 
 	initialized, err := p.pgm.IsInitialized()
 	if err != nil {
-		return nil, err
+		return pgState, err
 	}
 	if initialized {
 		pgParameters, err := p.pgm.GetConfigFilePGParameters()


cmd/keeper/cmd/keeper.go
```diff
@@ -641,7 +640,7 @@ func (p *PostgresKeeper) GetPGState(pctx context.Context) (*cluster.PostgresStat
 
 	initialized, err := p.pgm.IsInitialized()
 	if err != nil {
-		return nil, err
+		return pgState, err
```
sgotti: We could just do as the rest of the function, return the current pgstate (that will have healthy set to false) without returning an error. For the same reason we should change the signature of the function removing the error return since it will never return an error:

```
func (p *PostgresKeeper) GetPGState(pctx context.Context) *cluster.PostgresState
```

cmd/keeper/cmd/keeper.go
```diff
@@ -558,7 +558,6 @@ func (p *PostgresKeeper) updatePGState(pctx context.Context) {
 	pgState, err := p.GetPGState(pctx)
 	if err != nil {
 		log.Errorw("failed to get pg state", zap.Error(err))
-		return
```
sgotti: For the other comment just remote the error check since GetPGState will become:
```
func (p *PostgresKeeper) GetPGState(pctx context.Context) *cluster.PostgresState
```

cmd/keeper/cmd/keeper.go
maksm90: OK, done below. But I'm confused that the function `GetPGState` is called in [stage of node resync](https://github.com/sorintlab/stolon/blob/master/cmd/keeper/cmd/keeper.go#L1279) and returned error interrupts the resync process obviously. How to deal with this issue? For now, I doesn't interrupt the resync process anyway.

maksm90: @sgotti  I have reverted the last commit and leave the first version

sgotti:
@maksm90  Thanks. Can you please squash in a single commit so I can merge this PR?

maksm90:
> Can you please squash in a single commit so I can merge this PR?

Done

sgotti:
@maksm90 Thanks a lot! Merging.

maksm90:
> @maksm90 Thanks a lot! Merging.

Thanks too!

PR #639, opened 2019-05-02
keeper Prometheus metrics

lawrencejones:
Add a collection of Prometheus metrics to the keeper. The metrics are
aimed to expose errors in the keeper sync loop, providing enough
visibility to detect when the sync is failing (and some insight into
why).

---

Hey stolon maintainers! I want to make this as easy as possible to review, so I'm going to try and write everything you need up-front here. If you have any questions just drop a comment and I'll do my best to answer them.

We're currently moving our Postgres cluster to use stolon, and in doing so we've adapted some tooling we already had to enable zero-downtime planned failover. The tool that allows this is called [stolon-pgbouncer](https://github.com/gocardless/stolon-pgbouncer) which explains its purpose and aims in the readme.

stolon-pgbouncer has native Prometheus metrics and bundles with it some Prometheus alerts and dashboards. If you clone this repo and run `docker-compose up`, then navigate to http://localhost:3000 (login as admin/admin, skip password) you'll see two dashboards, one for the stolon-pgbouncer services and another for stolon-keepers.

The master branch of stolon-pgbouncer will boot with a compiled stolon-keeper binary from this PR. This enables us to scrape and display the stolon-keeper metrics in a dashboard we bundle with stolon-pgbouncer (PR that introduces this is [here](https://github.com/gocardless/stolon-pgbouncer/pull/29)) which looks like this:

image

It's meant to be a one-stop-shop for keeper state, while the [alerts](https://github.com/gocardless/stolon-pgbouncer/blob/master/docker/observability/prometheus/rules.yml#L146-L179) we've defined on these metrics aim to capture whenever keepers are misbehaving or Postgres is pending a restart.

We'd love to get these metrics upstreamed to benefit all stolon users, as well as obviously making our lives easier by avoiding maintaining a fork!

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sorintlab/stolon/internal/common"
 	"github.com/sorintlab/stolon/internal/store"
 	"github.com/sorintlab/stolon/internal/util"
@@ -75,6 +76,24 @@ func AddCommonFlags(cmd *cobra.Command, cfg *CommonConfig) {
 	}
 }
 
+var (
+	// clusterIdentifier provides a Prometheus metric that should uniquely identify the
+	// cluster that any stolon component is associated with. Users can then join between
+	// various metric series for the same cluster without making assumptions about service
+	// discovery labels.
+	clusterIdentifier = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "stolon_cluster_identifier",
+			Help: "Set to 1, is labelled with the cluster_name",
+		},
+		[]string{"cluster_name"},
+	)
+)
+
+func init() {
+	prometheus.MustRegister(clusterIdentifier)
+}
+
 func CheckCommonConfig(cfg *CommonConfig) error {
 	if cfg.ClusterName == "" {
 		return fmt.Errorf("cluster name required")
@@ -102,6 +121,13 @@ func CheckCommonConfig(cfg *CommonConfig) error {
 	}
 
 	return nil
+}
+
+// SetMetrics should be called by any stolon component that outputs application metrics.
+// It sets the clusterIdentifier metric, which is key to joining across all the other
+// metric series.
+func SetMetrics(cfg *CommonConfig) {
+	clusterIdentifier.WithLabelValues(cfg.ClusterName).Set(1)
 }
 
 func IsColorLoggerEnable(cmd *cobra.Command, cfg *CommonConfig) bool {
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -751,6 +751,10 @@ func (p *PostgresKeeper) Start(ctx context.Context) {
 	updatePGStateTimerCh := time.NewTimer(0).C
 	updateKeeperInfoTimerCh := time.NewTimer(0).C
 	for true {
+		// The sleepInterval can be updated during normal execution. Ensure we regularly
+		// refresh the metric to account for those changes.
+		sleepInterval.Set(float64(p.sleepInterval / time.Second))
+
 		select {
 		case <-ctx.Done():
 			log.Debugw("stopping stolon keeper")
@@ -966,6 +970,12 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 		log.Errorw("clusterdata validation failed", zap.Error(err))
 		return
 	}
+
+	// Mark that the clusterdata we've received is valid. We'll use this metric to detect
+	// when our store is failing to serve a valid clusterdata, so it's important we only
+	// update the metric here.
+	clusterdataLastValidUpdateSeconds.SetToCurrentTime()
+
 	if cd.Cluster != nil {
 		p.sleepInterval = cd.Cluster.DefSpec().SleepInterval.Duration
 		p.requestTimeout = cd.Cluster.DefSpec().RequestTimeout.Duration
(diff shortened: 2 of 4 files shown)


cmd/common.go
```diff
@@ -75,6 +76,24 @@ func AddCommonFlags(cmd *cobra.Command, cfg *CommonConfig) {
 	}
 }
 
+var (
+	// clusterIdentifier provides a Prometheus metric that should uniquely identify the
+	// cluster that any stolon component is associated with. Users can then join between
+	// various metric series for the same cluster without making assumptions about service
+	// discovery labels.
+	clusterIdentifier = prometheus.NewGaugeVec(
```
lawrencejones: This is the metric we (GoCardless) have been using to join various time series together. By having all components report this in a consistent fashion (see [here](https://github.com/gocardless/stolon-pgbouncer/blob/master/cmd/stolon-pgbouncer/main.go#L132-L138) for the stolon-pgbouncer definition) it becomes possible to join series from totally different process/infrastructures on the cluster_name and store_prefix labels.

cmd/keeper/cmd/keeper.go
```diff
@@ -1391,6 +1480,10 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
 	targetRole := db.Spec.Role
 	log.Debugw("target role", "targetRole", string(targetRole))
 
+	// Set metrics to power alerts about mismatched roles
```
lawrencejones: These metrics are particularly useful to power an up-to-date view of keeper state. It looks like this in a dashboard:

image

cmd/keeper/cmd/keeper.go
```diff
@@ -1770,6 +1875,7 @@ func (p *PostgresKeeper) generateHBA(cd *cluster.ClusterData, db *cluster.DB, on
 func sigHandler(sigs chan os.Signal, cancel context.CancelFunc) {
 	s := <-sigs
 	log.Debugw("got signal", "signal", s)
+	shutdownSeconds.SetToCurrentTime()
```
lawrencejones: Useful for detecting a pending shutdown.

internal/common/common.go
```diff
@@ -42,6 +42,13 @@ const (
 	RoleStandby   Role = "standby"
 )
 
+// Roles enumerates all possible Role values
```
lawrencejones: This list needs maintaining to ensure we can set all possible label values to 0 on stolon_keeper_{local,target}_role before setting our active metric to be 1. This becomes clearer when you see the output of a single instances metrics:

image

sgotti:
@lawrencejones Thanks a lot for the PR and the detailed description. I'll review it in the next days.

lawrencejones:
Hey @sgotti, I just realised the semaphoreci tests have failed here. Is that normal or should I debug that?

cmd/common.go
sgotti: Ideally only the clustername should be neeed (the prefix isn't used when running stolon inside k8s). Obviously people could setup multiple instances using the same cluster name and prefix using different store instances. But I think people that use multiple stolon instances should also use different cluster names if they want to distinguish them inside prometheus.

Another solution will be to add another "clusteruid" option but I think this is redundanto and should be covered by the "clustername"

cmd/keeper/cmd/keeper.go
```diff
@@ -943,6 +948,84 @@ func (p *PostgresKeeper) refreshReplicationSlots(cd *cluster.ClusterData, db *cl
...
+	setRole(targetRoleGauge, nil)
+	prometheus.MustRegister(localRoleGauge)
+	setRole(localRoleGauge, nil)
+	prometheus.MustRegister(needsReloadGauge)
+	prometheus.MustRegister(needsRestartGauge)
+	prometheus.MustRegister(lastSyncSuccessSeconds)
+	prometheus.MustRegister(sleepInterval)
+	prometheus.MustRegister(shutdownSeconds)
+}
+
```
sgotti: The keeper.go file is just quite big. Can you please move this block inside `cmd/keeper/cmd/prometheus.go` ?

lawrencejones:
@AMyltsev we're definitely interested in doing this but wanted to start with the keepers first as the most impactful area to add visibility. We had keepers failing to start or being stuck in an initialisation step that we'd otherwise not hear about- that situation wasn't acceptable for us!

If you're interested in tackling the proxy/sentinel metrics yourself as a follow-up to this PR then I'd happily help review? I think this PR can be merged independently of the other components though, and it will benefit the following work by setting an example of how you might approach them.

cmd/common.go
```diff
@@ -75,6 +76,24 @@ func AddCommonFlags(cmd *cobra.Command, cfg *CommonConfig) {
...
+		prometheus.GaugeOpts{
+			Name: "stolon_cluster_identifier",
+			Help: "Set to 1, is labelled with store_prefix and cluster_name",
```
sgotti: "Set to 1, is labelled with the cluster_name",

cmd/keeper/cmd/metrics.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2017 Sorint.lab
```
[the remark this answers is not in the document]
sgotti: :+1:

cmd/common.go
lawrencejones: 🤦‍♂

lawrencejones:
I think all changes are applied now @sgotti, and I've freshly rebased against master. Thanks for all your help!

lawrencejones:
Hey @sgotti, I think everything is fixed here. Don't mean to bother you but if we're good, merging this would make things much easier for me to manage follow-up PRs for sentinel/etc metrics.

Is there anything else you'd like here?

sgotti:
@lawrencejones LGTM! Mergine.

Issue #314, opened 2017-07-13
Configurable proxy*interval


LordFPL:
Hello,

Can it be possible to add a way to configure : 
```
DefaultProxyCheckInterval   = 5 * time.Second
DefaultProxyTimeoutInterval = 15 * time.Second
```
Or there is a reason to keep theses values ?

sgotti:
@LordFPL Now they are hardcoded.

May I ask you the reason why you want to change them? (timeouts?).

They could probably be made configurable since #310 has been merged but we should add some checks (timeout > checkinterval) and put an upper bound since increasing them to too big values could increase the failover time.

LordFPL:
Only because during my tests of failover, i modify "sleepInterval" to 1s... and was surprised to see proxy chek always on 5s because documentation say : 
```
sleepInterval | interval to wait before next check (for every component: keeper, sentinel, proxy). | no | string (duration) | 5s
```
I'm trying to see how i can reduce failover time.
Not a very big problem, 5s is already nice :)

sgotti:
@LordFPL yeah, the doc is currently not right (we changed the proxy logic without fixing the doc).


> I'm trying to see how i can reduce failover time.

As a warning, now we are not checking all the values and their possible interaction (for example setting `failInterval` < `sleepInterval` will probably cause the keepers to be marked as failed if they're missing only a single check).

Also lowering the various intervals to too low ones can cause some unwanted failovers (for example if the machine is under load).

trmfreitas:
I am doing this change but I am not validating values interaction.

sgotti:
Implemented in #756.

Issue #588, opened 2018-11-26
Database disconnection acquired. Context deadline exceeded.


catroot:
Database disconnection acquired... the proxy logs tells that:
```
2018-11-26T19:43:51.698Z	INFO	cmd/proxy.go:302	check function error	{"error": "failed to update proxyInfo: context deadline exceeded"}
....
2018-11-26T19:44:16.407Z	INFO	cmd/proxy.go:249	not proxying to master address since we aren't in the enabled proxies list	{"address": "10.0.2.65:5432"},

```
Any way to handle context deadline?

sgotti:
@catroot this is normal if your underlying store (etcd) cannot answer to request (usually overloaded). You should check why this is happening.

sgotti:
@catroot currently it's hardcoded to 5 seconds and the proxy will timeout in 15 second. Which is IMHO already quite a lot. We'll happily accept PR to make it configurable.

catroot:
Yes, please. I`m using stolon behind docker and application does a heavy transactions, (db dumping for example) and kicks off the connection due to this parameter.
For my sandbox i have all of them run on the one physical machine and heavy load hangs other things.

trmfreitas:
@catroot I am doing this change, also had the same problem you did

sgotti:
Configurable proxyCheckInterval` and `proxyTimeout` have been implemented in #756. Etcd store timeout is hardcoded to 5 second and a store taking more than 5 second to answer is really slow and should be fixed.

PR #643, opened 2019-05-16
Configurable proxy check interval and request timeout

trmfreitas:
Hi!

I added configuration of request timeout and check interval for stolon-proxy.
Configured as arguments:
--check-interval-seconds <seconds> --request-timeout-seconds <seconds>

Also changed help and documentation.

This should fix:
#314

And help on:
#588 

Thanks.

Best regards,
Tiago Freitas


sgotti:
@trmfreitas Thanks for your PR. Unfortunately it's not possible to change them per proxy since this value is also used (and must be in sync) in the sentinel: https://github.com/sorintlab/stolon/blob/6003d65273d7891aa7ad066ac7d1002bb33ca43c/cmd/sentinel/cmd/sentinel.go#L343 or all the fence logic will break leading to bad things.

So it should be a cluster spec option.

trmfreitas:
So, should the value be the same in both places?
If it is, then I will try to move it to clusterSpec. What would be a better name for the parameter?

Thanks.

trmfreitas:
I am closing this PR and will open a new clean one

Issue #601, opened 2019-01-24
resyncronization via pg_rewind


glushakov:
Hi.
I want to discuss some issues with resyncing old master.
I spent a lot of time to investigate why pg_rewind not used in my case. 
will describe it:

1-st try:
1. Gracefull shutdown Master. Standby successfully promoted
2. wait for 30 seconds - starting ex-master. Watch the logs:

2019-01-23T14:44:32.656+0300    INFO    cmd/keeper.go:801       syncing using pg_rewind {"followedDB": "d8cc83a6", "keeper": "serv1"}
2019-01-23T14:44:32.656+0300    INFO    postgresql/postgresql.go:783    running pg_rewind
fetched file "global/pg_control", length 8192
source and target cluster are on the same timeline
no rewind required
2019-01-23T14:44:32.852+0300    INFO    postgresql/postgresql.go:307    starting database
2019-01-23 14:44:32.963 MSK [22526] LOG:  listening on IPv4 address "127.0.0.1", port 5432
2019-01-23 14:44:32.978 MSK [22526] LOG:  listening on IPv4 address "10.243.196.53", port 5432
2019-01-23 14:44:32.982 MSK [22526] LOG:  listening on Unix socket "/tmp/.s.PGSQL.5432"
2019-01-23 14:44:33.071 MSK [22526] LOG:  redirecting log output to logging collector process
2019-01-23 14:44:33.071 MSK [22526] HINT:  Future log output will appear in directory "log".
2019-01-23T14:44:33.754+0300    INFO    cmd/keeper.go:870       followed instance timeline forked before our current state      {"followedTimeline": 12, "followedXlogpos": 871782152576, "timeline": 11, "xlogpos": 871795620912}
2019-01-23T14:44:33.756+0300    INFO    postgresql/postgresql.go:372    stopping database
waiting for server to shut down.....2019-01-23T14:44:35.108+0300        ERROR   cmd/keeper.go:641       cannot get configured pg parameters     {"error": "pq: the database system is shutting down"}
..2019-01-23T14:44:37.611+0300  ERROR   cmd/keeper.go:641       cannot get configured pg parameters     {"error": "pq: the database system is shutting down"}
.. done
server stopped
2019-01-23T14:44:40.912+0300    INFO    cmd/keeper.go:827       syncing from followed db        {"followedDB": "d8cc83a6", "keeper": "serv1"}
[...]

PR #644, opened 2019-05-17
[WIP] Force checkpoint before rewind

lawrencejones:
**This needs follow-up integration tests and making the flag configurable. We're going to use this in our fork for the meantime and commit to improving things next week.**

Prior to running pg_rewind, force a checkpoint on the source database to
ensure the pg_control file used by pg_rewind is up-to-date. Failure to
do so can result in pg_rewind not detecting a timeline fork and exiting
with no action, leading stolon to fallback on pg_basebackup (which
itself will checkpoint) to ensure a resync is successful.


lawrencejones:
This actually addresses #601, though not via any of the remediations recommended in that issue.

Frank111:
> wal_keep_segments

how to change wal_keep_segments？

lawrencejones:
Hey @rearden-steel @frank111, I'm not sure how this is related to the wal keep segments GUC? This change is about recovering from forked timelines rather than having insufficient wal to perform a recovery.

> can I increase it manually in code

We've done so in our stolon fork as we also needed a higher value than 8. It's safe to do so, but I'd recommend pushing https://github.com/sorintlab/stolon/pull/591 so we can get this solved for everyone.

tests/integration/ha_test.go
```diff
@@ -1101,6 +1097,32 @@ func testTimelineFork(t *testing.T, syncRepl, usePgrewind bool) {
...
 
+	// If enabled, it should be possible to recover from forked timelines by using
+	// pg_rewind. It's possible that pg_rewind will fail and we'd still recover the cluster
+	// by falling-back to pg_basebackup.
+	//
+	// Most people who enable pg_rewind really don't want basebackup to happen, as it may
+	// take hours to recover when pg_rewind is <1m. Here we verify that the keeper
+	// successfully recovered the forked timeline via a rewind rather than basebackup.
+	if usePgrewind {
+		output := standbys[1].ReadOutput()
```
sgotti: This should already be covered by using Expect (see config_test.go)

tests/integration/utils.go
```diff
@@ -253,7 +255,9 @@ func (p *Process) start() error {
 	go func() {
 		scanner := bufio.NewScanner(pr)
 		for scanner.Scan() {
-			p.t.Logf("[%s %s]: %s", p.name, p.uid, scanner.Text())
```
sgotti: This should already be covered by using Expect (see config_test.go)

rearden-steel:
Hey @lawrencejones!
Is there any chance to make the requested changes?
We experience issues which should be fixed with this PR.

lawrencejones:
Hey @rearden-steel, I'm.midway through making this work. It needs integration tests which are what I've been working on, and to provides configurable variable that enables/disables checkpointing. 

I'm out at a conference today/tomorrow but this is right on my to-do list come Thursday. I'm going to push the WIP changes now so you can see them, and caveat this update by saying I'm explicitly ignoring Simone's advice until I've got the tests working (at which point I'll action them).

lawrencejones:
From where I got to last time, it's difficult to simulate a Postgres that takes a while to checkpoint after a promotion. While pg_ctl promote won't force a synchronous checkpoint it will schedule one, and in the test databases that scheduled checkpoint is almost instantaneous leading to the controlfile being updated immediately.

Hopefully you can wait a couple of days for me to get back to this?

rearden-steel:
> it's difficult to simulate a Postgres that takes a while to checkpoint after a promotion
I was able to reach such situation when doing manual failover while running pgbench (-c 100 -T 120 -j 20) on a cluster (with pre-generated database around 80 GB, pgbench -i -s 5000).
> Hopefully you can wait a couple of days for me to get back to this?
Sure! Thank you for your work!

tests/integration/utils.go
lawrencejones: Hey @sgotti I'm struggling to figure out how to use the expect interface to try matching against what I need here.

I need to verify that this string occurs after a particular place in the code (so flush the buffer, then run the assertion) and I need to perform multiple assertions for several strings that may appear out of order.

Can you help me figure out how to do this with Expect?

lawrencejones: Figured this out actually, thanks!

lawrencejones:
@sgotti @rearden-steel this is ready for review now.

Instead of adding this logic into an existing test, I've created another test with just two keepers that verifies we're safe from this specific race. The details of that test can be seen here: https://github.com/sorintlab/stolon/blob/bd61324616e4e501ec511e10f27e2c09fac8e130/tests/integration/ha_test.go#L1174-L1187

I've added some test helpers and I'm going to go through the diff and comment where things may not be totally clear.

Any questions, please do ask!

cmd/keeper/cmd/keeper.go
```diff
@@ -1284,19 +1286,18 @@ func (p *PostgresKeeper) postgresKeeperSM(pctx context.Context) {
...
-			// pg_rewind disabled. Will need to report this upstream.
-
-			// TODO(sgotti) The rewinded standby needs wal from the master
-			// starting from the common ancestor, if they aren't available the
-			// instance will keep waiting for them, now we assume that if the
-			// instance isn't ready after the start timeout, it's waiting for
-			// wals and we'll force a full resync.
-			// We have to find a better way to detect if a standby is waiting
-			// for unavailable wals.
+			// TODO(sgotti) pg_rewind considers databases on the same timeline as in sync and
```
lawrencejones: I edited this comment as part of an unrelated change and line-wrapped it to match the rest of the file. Can leave it or remove it as makes most sense!

tests/integration/ha_test.go
```diff
@@ -124,18 +124,21 @@ func TestInitWithMultipleKeepers(t *testing.T) {
 }
 
 func setupServers(t *testing.T, clusterName, dir string, numKeepers, numSentinels uint8, syncRepl bool, usePgrewind bool, primaryKeeper *TestKeeper) (testKeepers, testSentinels, *TestProxy, *TestStore) {
-	var initialClusterSpec *cluster.ClusterSpec
+	var initialClusterSpec = &cluster.ClusterSpec{
```
lawrencejones: Small refactor here to put the common ClusterSpec up top, then assign the overrides. It might make sense to use something like https://github.com/imdario/mergo in future, so we can merge optional overrides provided in this `setupServers` helper.

tests/integration/ha_test.go
```diff
@@ -124,18 +124,21 @@ func TestInitWithMultipleKeepers(t *testing.T) {
...
+		SleepInterval:          &cluster.Duration{Duration: 2 * time.Second},
+		FailInterval:           &cluster.Duration{Duration: 5 * time.Second},
+		ConvergenceTimeout:     &cluster.Duration{Duration: 30 * time.Second},
+		MaxStandbyLag:          cluster.Uint32P(50 * 1024), // limit lag to 50kiB
+		SynchronousReplication: cluster.BoolP(syncRepl),
+		PGParameters:           defaultPGParameters,
+		// If we want to pg_rewind then also checkpoint beforehand, otherwise we may fail due
+		// to uncheckpoint'ed timeline changes that have us silently fallback to basebackup.
+		UsePgrewind:              cluster.BoolP(usePgrewind),
+		CheckpointBeforePgrewind: cluster.BoolP(usePgrewind),
```
lawrencejones: This is open for questioning: should we assign this by default? It doesn't harm us now but we may want to test when the parameter is *not* set (though I think this is unlikely).

Perhaps ok for now, until we have an issue?

tests/integration/ha_test.go
```diff
@@ -256,6 +270,32 @@ func waitLines(t *testing.T, q Querier, num int, timeout time.Duration) error {
 	return fmt.Errorf("timeout waiting for %d lines, got: %d", num, c)
 }
 
+// waitForLine waits until a row with the expected ID appears in table01
+func waitForLine(t *testing.T, q Querier, expected int, timeout time.Duration) error {
```
lawrencejones: This was what I expected `waitForLines` to do! Tripped myself up by assuming its behaviour. Two small things here:

- I've tried to comment my new test helpers to make it clear why you might use them, and how they behave. I would have found it really useful to have comments like these on the ones that already exist when getting familiar with the test suite, though I totally understand when this is mostly a one man operation why they wouldn't be there
- I'm a bit worried `waitLines` might be too loose an assertion in some tests. It could be that we're passing tests when a specific token value never replicated, while a subsequent one did. Not sure on this but I'd normally avoid this type of assertion in favour of something more specific, like this `waitForLine`

tests/integration/utils.go
```diff
@@ -52,7 +52,11 @@ const (
 )
 
 var (
-	defaultPGParameters = cluster.PGParameters{"log_destination": "stderr", "logging_collector": "false"}
+	defaultPGParameters = cluster.PGParameters{
+		"log_destination":   "stderr",
+		"logging_collector": "false",
+		"log_checkpoints":   "on",
```
lawrencejones: Can't see why we wouldn't have this on. It doesn't hurt us and can be extremely useful.

tests/integration/utils.go
```diff
@@ -438,6 +442,42 @@ func (tk *TestKeeper) PGDataVersion() (int, int, error) {
 	return pg.ParseVersion(version)
 }
 
+// GetPGControldata calls pg_controldata for the Postgres database directory. This can be
+// useful in debugging database state, especially when comparing what was successfully
+// checkpointed.
+func (tk *TestKeeper) GetPGControldata() string {
```
lawrencejones: Unused but was useful while debugging. Can remove if we want, or leave for future use.

tests/integration/ha_test.go
sgotti: I think the result is the same just with different logic:

What the tests really want to check is that all the transactions are applied. Waiting for the number of rows in an append only table is a way to do this.

If you only append rows to a table you know what the expected total number should be. So I think waiting for the total number of rows is enough. Waiting for the last row you inserted is another way for doing the same thing. Or am I missing something?

I'll stick to just one of them to not add additional confusion the the integration tests (if you really prefer waiting for the last row feel free to create a PR that replaces it)




P.S. The real issue is the function name that will require a good s/line/row/ :smile:

lawrencejones: In this test, we're expecting to lose data. So we do three writes of 1, 2 and 3, but the final database should only ever get 1 and 3. 

If I use the count of lines then I'd have to expect there to be 2 lines, but that will pass even if I never received the correct write and only have 1 and 2 in the final standby (while the primary node has 1 and 3). 

Does that make any sense?

sgotti:
@lawrencejones Thanks for your PR.

Beside the inline comments, one thing that I'd like to avoid at all costs is issuing "write" commands from one keeper to another instance since it looks quite dirty (against the fact that the postgres should be managed only by its keeper and, in this case, every resync will issue an unneeded checkpoint).

So, shouldn't be possible to just issue a checkpoint after a promote by the primary keeper?

The main issue I see here is that there'll be a window between issuing a promote and finishing the checkpoint where the keeper already reported its state to the sentinel and it has updated the old primary follow config.
One possible solution to fix this and also other realted issues will be to report to the sentinel the current checkpoint status after promote or the checkpoint timeline (new field in cluster.go `DBStatus`) so the sentinel will set the old primary `followConfig` only when it's sure it's in good state for a pg_rewind. 

P.S. the same logic could be expanded to also report the current replication slots so the sentinel can avoid similar issues (and remove the use of the `--slot` command in pg_basebackup that is another hacky thing).

lawrencejones:
Hey @sgotti! I've been rubbish here: PRs were left hanging while I went on holiday 🏖 

I'm going to reply to each bit of your review separately. Hopefully we can reason our way to a conclusion!

> Beside the inline comments, one thing that I'd like to avoid at all costs is issuing "write" commands from one keeper to another instance since it looks quite dirty (against the fact that the postgres should be managed only by its keeper and, in this case, every resync will issue an unneeded checkpoint).
>
> So, shouldn't be possible to just issue a checkpoint after a promote by the primary keeper?

Postgres will automatically begin a checkpoint whenever a promote happens, but it will do so in a spread fashion (ie, using all the checkpointing configuration parameters to ensure it throttles the checkpointing). If we were to get the keeper to issue a checkpoint itself after promotion (optionally, via a similar configuration parameter as this PR) then we'd force the master to checkpoint faster, which is good but...

At the same time as the master checkpoints, our slaves will attempt to resync against the master. Until the checkpoint completes the pg_rewind could fail, and we'll fallback to pg_basebackup, which is what we really want to avoid.

If we really want to avoid this, then a possible solution would be to provide this checkpoint-after-promote functionality we've been discussing, but additionally have keepers retry pg_rewind for a specific time threshold. A user could then set the retry timeout to be 5m, for example, and enable checkpoint-after-promote to ensure the rewind will be retried after the checkpoint is expected to be completed.

> The main issue I see here is that there'll be a window between issuing a promote and finishing the checkpoint where the keeper already reported its state to the sentinel and it has updated the old primary follow config.
> 
> One possible solution to fix this and also other realted issues will be to report to the sentinel the current checkpoint status after promote or the checkpoint timeline (new field in cluster.go DBStatus) so the sentinel will set the old primary followConfig only when it's sure it's in good state for a pg_rewind.

It's quite sad for me that we have to do this, if you know what I mean? pg_basebackup has an option to wait for a checkpoint to complete before starting, so it's quite sad to me that pg_rewind doesn't also have this option. Perhaps we can simulate it by creating a backup label, as that appears to be what pg_basebackup does internally? Do you have any thoughts on this?

lawrencejones:
@sgotti I know you're probably busy on a lot of other things right now, but would really appreciate your feedback for this change. It matters quite a lot to us that rewinds are working, and this change makes that the case.

If we want to take another path that's great, but we'll need to know what it is before we can get implementing it!

sgotti:
@lawrencejones Also if theoretically I'd prefer a solution that doesn't require doing operations from a keeper to another, in the end this is already done when doing a pg_basebackup or a pg_rewind.

So I don't want to stop this PR trying to find a cleanliness that's not really possible. Lets go ahead with your solution! (in the end we are just doing what pg_rewind should do: a checkpoint inside pg_rewind like pg_basebackup already does)

P.S. We should report this upstream if not already done.

internal/postgresql/postgresql.go
```diff
@@ -786,6 +786,32 @@ func (p *Manager) SyncFromFollowedPGRewind(followedConnParams ConnParams, passwo
...
+	// checkpointed, the pg_control file will tell us we're both on the same timeline (1)
+	// and pg_rewind will exit without performing any action.
+	//
+	// If we checkpoint before invoking pg_rewind we will avoid this problem, at the slight
+	// cost of forcing a checkpoint on a newly promoted node, which might hurt performance.
+	// We (GoCardless) can't afford this, so we take the performance penalty to avoid hours
+	// of downtime.
+	if forceCheckpoint {
+		log.Infow("issuing checkpoint on primary")
+		psqlName := filepath.Join(p.pgBinPath, "psql")
```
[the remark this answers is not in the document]
lawrencejones: I was doing this to ensure the way we connect to the Postgres for pg_rewind, pg_basebackup and for issuing a checkpoint was consistent. psql should behave exactly the same as rewind/basebackup, whereas connecting from within Go could be subtly different in many ways.

Does that make sense, or do you think we should try constructing a Go connection?

lawrencejones:
Hey @sgotti!

> So I don't want to stop this PR trying to find a cleanliness that's not really possible. Lets go ahead with your solution! (in the end we are just doing what pg_rewind should do: a checkpoint inside pg_rewind like pg_basebackup already does)

Makes sense to me :)

I've adjusted the doc comment to make it clearer we'll be hitting the primary. All your other review comments were questions/have been addressed, I think.

Can I get another look? Thanks.

viggy28:
Hi Team, wondering any updates on this? Anything I can help with to get this merged?

sgotti:
@viggy28 If no one is going to work in this PR (@lawrencejones) feel free to adopt and update it.

lawrencejones:
Hey @viggy28, if you want to take this over then please let me know. I’m not working on stolon things right now so if you need this, I’ll be happy to hand it over!

viggy28:
Sure @lawrencejones. Looks like you have done most of the work. Thanks much.

Trying to understand what is needed.

1. https://github.com/sorintlab/stolon/pull/644/files#r293416129
2. https://github.com/sorintlab/stolon/pull/644#pullrequestreview-249368918 is unresolved but @sgotti also mentioned here https://github.com/sorintlab/stolon/pull/644#issuecomment-524313267 that shouldn't be a blocker. which means can we resolve that?
3. Looks like the build is failing. 

Anything else I am missing?

internal/postgresql/postgresql.go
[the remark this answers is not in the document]
viggy28: https://gist.github.com/viggy28/954ff01cd3c29d317834a5c25951a1cd

@lawrencejones and @sgotti does this look okay? (since sslmode prefer is not applicable for lib/pq I have to replace that based on the SSL settings on the cluster).

pirateunclejack:
Dear contributors , 
We have a stolon cluster which is such big to cost hours to recovery with basebackup.
I apply this change to latest master and run on our test environment . the standby keepers keep failed after pg_rewind done : 
```

pg_rewind: Done!
2024-02-27T09:24:12.615Z	INFO	cmd/keeper.go:928	syncing using pg_rewind	{"masterDB": "1f5bbc57", "keeper": "keeper0"}
2024-02-27T09:24:12.636Z	INFO	postgresql/postgresql.go:921	issuing checkpoint on primary
CHECKPOINT
2024-02-27T09:24:12.993Z	INFO	postgresql/postgresql.go:932	running pg_rewind
2024-02-27T09:24:12.994Z	DEBUG	postgresql/postgresql.go:936	execing cmd	{"cmd": "/usr/lib/postgresql/14/bin/pg_rewind --debug -D /stolon-data/postgres --source-server=application_name=stolon_7805b990 dbname=zabbix host=172.16.250.153 options=-c\\ synchronous_commit=off password=******************** port=5432 sslmode=prefer user=stolon"}
pg_rewind: executing "/usr/lib/postgresql/14/bin/postgres" for target server to complete crash recovery
2024-02-27 09:24:13.068 GMT [1590] LOG:  skipping missing configuration file "/stolon-data/postgres/postgresql.auto.conf"
2024-02-27T09:24:13.173Z	ERROR	cmd/keeper.go:736	cannot get configured pg parameters	{"error": "dial unix /tmp/.s.PGSQL.5432: connect: no such file or directory"}
2024-02-27T09:24:15.675Z	ERROR	cmd/keeper.go:736	cannot get configured pg parameters	{"error": "dial unix /tmp/.s.PGSQL.5432: connect: no such file or directory"}
2024-02-27 09:24:16.349 UTC [1590] LOG:  database system was interrupted while in recovery at log time 2024-02-27 09:23:11 UTC
2024-02-27 09:24:16.349 UTC [1590] HINT:  If this has occurred more than once some data might be corrupted and you might need to choose an earlier recovery target.
2024-02-27 09:24:16.455 UTC [1590] LOG:  redo starts at B/280161F8
2024-02-27 09:24:16.556 UTC [1590] LOG:  invalid record length at B/28182E08: wanted 24, got 0
2024-02-27 09:24:16.556 UTC [1590] LOG:  redo done at B/28182DE0 system usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.10 s

PostgreSQL stand-alone backend 14.0 (Debian 14.0-1.pgdg110+1)
2024-02-27T09:24:17.548Z	DEBUG	postgresql/postgresql.go:625	execing cmd	{"cmd": "/usr/lib/postgresql/14/bin/postgres -V"}
2024-02-27T09:24:17.561Z	DEBUG	cmd/keeper.go:815	pgstate dump: (*cluster.PostgresState)(0xc0008f84d0)({
 UID: (string) (len=8) "7805b990",
 Generation: (int64) 0,
 ListenAddress: (string) (len=14) "172.16.115.237",
 Port: (string) (len=4) "5432",
 Healthy: (bool) false,
 SystemID: (string) "",
 TimelineID: (uint64) 0,
 XLogPos: (uint64) 0,
 TimelinesHistory: (cluster.PostgresTimelinesHistory) <nil>,
 PGParameters: (common.Parameters) <nil>,
 SynchronousStandbys: ([]string) <nil>,
 OlderWalFile: (string) ""
})

2024-02-27T09:24:18.177Z	ERROR	cmd/keeper.go:736	cannot get configured pg parameters	{"error": "dial unix /tmp/.s.PGSQL.5432: connect: no such file or directory"}
pg_rewind: fetched file "global/pg_control", length 8192
pg_rewind: source and target cluster are on the same timeline
pg_rewind: no rewind required
2024-02-27T09:24:18.222Z	DEBUG	postgresql/postgresql.go:625	execing cmd	{"cmd": "/usr/lib/postgresql/14/bin/postgres -V"}
2024-02-27T09:24:18.324Z	INFO	postgresql/postgresql.go:844	writing standby signal file
2024-02-27T09:24:18.368Z	INFO	postgresql/postgresql.go:319	starting database
2024-02-27T09:24:18.368Z	DEBUG	postgresql/postgresql.go:323	execing cmd	{"cmd": "/usr/lib/postgresql/14/bin/postgres -D /stolon-data/postgres -c unix_socket_directories=/tmp"}
2024-02-27T09:24:20.682Z	ERROR	cmd/keeper.go:736	cannot get configured pg parameters	{"error": "dial unix /tmp/.s.PGSQL.5432: connect: no such file or directory"}
2024-02-27 09:24:21.540 UTC [1593] LOG:  starting PostgreSQL 14.0 (Debian 14.0-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
2024-02-27 09:24:21.540 UTC [1593] LOG:  listening on IPv4 address "172.16.115.237", port 5432
2024-02-27 09:24:21.589 UTC [1593] LOG:  listening on Unix socket "/tmp/.s.PGSQL.5432"
2024-02-27 09:24:22.064 UTC [1594] LOG:  database system was shut down at 2024-02-27 09:24:17 UTC
2024-02-27 09:24:22.064 UTC [1595] FATAL:  the database system is starting up
2024-02-27 09:24:22.085 UTC [1594] LOG:  entering standby mode
2024-02-27 09:24:22.221 UTC [1594] LOG:  invalid record length at B/28182EF8: wanted 24, got 0
2024-02-27 09:24:22.245 UTC [1598] LOG:  started streaming WAL from primary at B/28000000 on timeline 31
2024-02-27 09:24:22.267 UTC [1599] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:22.267 UTC [1599] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27 09:24:22.271 UTC [1594] LOG:  invalid resource manager ID 48 at B/28182EF8
2024-02-27 09:24:22.279 UTC [1598] FATAL:  terminating walreceiver process due to administrator command
2024-02-27 09:24:22.279 UTC [1594] LOG:  invalid resource manager ID 48 at B/28182EF8
2024-02-27 09:24:22.280 UTC [1594] LOG:  invalid resource manager ID 48 at B/28182EF8
2024-02-27 09:24:22.469 UTC [1601] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:22.469 UTC [1601] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27T09:24:22.583Z	DEBUG	postgresql/postgresql.go:625	execing cmd	{"cmd": "/usr/lib/postgresql/14/bin/postgres -V"}
2024-02-27T09:24:22.593Z	DEBUG	cmd/keeper.go:815	pgstate dump: (*cluster.PostgresState)(0xc0006ac630)({
 UID: (string) (len=8) "7805b990",
 Generation: (int64) 0,
 ListenAddress: (string) (len=14) "172.16.115.237",
 Port: (string) (len=4) "5432",
 Healthy: (bool) false,
 SystemID: (string) "",
 TimelineID: (uint64) 0,
 XLogPos: (uint64) 0,
 TimelinesHistory: (cluster.PostgresTimelinesHistory) <nil>,
 PGParameters: (common.Parameters) <nil>,
 SynchronousStandbys: ([]string) <nil>,
 OlderWalFile: (string) ""
})

2024-02-27 09:24:22.671 UTC [1603] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:22.671 UTC [1603] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27 09:24:22.873 UTC [1604] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:22.873 UTC [1604] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27 09:24:23.094 UTC [1605] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:23.094 UTC [1605] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27 09:24:23.185 UTC [1606] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:23.185 UTC [1606] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27T09:24:23.185Z	ERROR	cmd/keeper.go:736	cannot get configured pg parameters	{"error": "pq: the database system is not yet accepting connections"}
2024-02-27 09:24:23.297 UTC [1607] FATAL:  the database system is not yet accepting connections
2024-02-27 09:24:23.297 UTC [1607] DETAIL:  Consistent recovery state has not been yet reached.
2024-02-27 09:24:23.499 UTC [1608] FATAL:  the database system is not yet accepting connections
```

Is there any body still work on this issue ?

PR #645, opened 2019-05-21
Make DBWaitReadyTimeout configurable

hmac:
This value is currently hardcoded to 60 seconds, and it is the time that
stolon will wait for Postgres to boot. In some circumstances (e.g. if
there is a lot of WAL to recover) Postgres can take longer than 60
seconds to boot. Allowing this value to be configured lets users cater
for that scenario. We do this by adding it to the cluster spec, so that
this value is consistent across all nodes in the cluster.


doc/cluster_spec.md
```diff
@@ -38,6 +38,7 @@ Some options in a running cluster specification can be changed to update the des
...
 | pgHBA                     | a list containing additional pg_hba.conf entries. They will be added to the pg_hba.conf generated by stolon. **NOTE**: these lines aren't validated so if some of them are wrong postgres will refuse to start or, on reload, will log a warning and ignore the updated pg_hba.conf file                                                                                                                                                                                          | no                        | []string          | null. Will use the default behiavior of accepting connections from all hosts for all dbs and users with md5 password authentication                                                   |
 | automaticPgRestart        | restart postgres automatically after changing the pgParameters that requires restart. Refer `pending_restart` in [pg_settings](https://www.postgresql.org/docs/9.5/static/view-pg-settings.html)    | no | bool | false |
+| dbWaitReadyTimeout        | Time to wait for the database to become ready after starting.  Increase this value if your Postgres takes longer to boot, e.g. because it has to recover a lot of WAL.
```
sgotti: The `Required`, `Type`, `Default` columns values are missing

internal/cluster/cluster.go
```diff
@@ -222,6 +222,8 @@ type ClusterSpec struct {
 	InitTimeout *Duration `json:"initTimeout,omitempty"`
 	// Interval to wait for a db to be synced with a master
 	SyncTimeout *Duration `json:"syncTimeout,omitempty"`
+	// Interval to wait for a db to boot and become ready
+	DBWaitReadyTimeout *Duration `json:"dbWaitReadyTimeout,omitempty"`
```
sgotti: s.DBWaitReasyTimeout is not used by the keeper that will continue to use DefaultDBWaitReadyTimeout. You should also change cmd/keeper/cmd/keeper.go

hmac: 🤦‍♂ sorry about that! Cherry picked one commit and missed the other! Updated.

sgotti:
@hmac Thanks a lot! Merging.

hmac:
@sgotti brilliant, thank you!

PR #648, opened 2019-05-24
Add badges in README

dineshba:
Added three badges in readme:
- Go report card
- Total downloads of latest version
- Latest version number


sgotti:
@dineshba Let's collect some badges. Merging. Thanks!

PR #651, opened 2019-05-25
Store only last 2 postgres timeline histories

### LINKED ISSUE #629: Clusterdata too large to store in consul
### Submission type

 - Request for enhancement (RFE)

### Environment

two hosts, proxy+keeper+sentinel on each host
store backend: consul

#### Stolon version

v0.13.0-pg10

All stolon services running in docker container.
I am testing stolon by randomly killing some stolon component(s).

### Expected behaviour you didn't see

One keeper failed to restore after `docker restart` command

### Unexpected behaviour you saw

I noticed stolon will store keeper recovery history in clusterdata,  
but on the 3205th recovery process,  
clusterdata size exceeds k/v store value size limit (512KB).  
Then stolon failed to write clusterdata to consul.  

`timelinesHistory` is full of:
```
{
    "timelineID": 1,
    "switchPoint": 301989984,
    "reason": "no recovery target specified"
},
```

### Enhancement Description
I tried to remove `timelinesHistory` entries in clusterdata,  
but it didn't work, looks like `timelineID`s are used elsewhere.  

Are there any methods to clean `timelinesHistory` in clusterdata?

dineshba:
Fixes #629 

We have made changes to **store last N (MaxPostgresTimelinesHistory) timeline histories**. We have set the default value as `1000`.

We introduced an interface `PGManager` to mock `GetTimelinesHistory` method in unit test.

Question:
- Does it make sense to make `MaxPostgresTimelinesHistory` configurable through `clusterspec`?
- What would be the suitable default value for `MaxPostgresTimelinesHistory`?


sgotti:
@dineshba Thanks for the PR!

TimeLineHistory is used to check that the current db is not diverged and is used in function `isDifferentTimelineBranch` in both keeper.go and sentinel.go (they are quite identical)

Looking at these functions logic we could theoretically just keep the last two timeline histories.

So we could avoid having an user defined (and unclear from the user point of view) parameter and just keep the last N (where N could be just the last 2 or be conservative and choose something like 10...)

dineshba:
Make sense @sgotti. It couldn't be user defined parameter. We can make `N as 2`. Will make the changes and update the PR.

dineshba:
@sgotti Updated the PR with `maxPostgresTimelinesHistory = 2`(made it keeper.go level const declaration). Squashed and did force push to have clean history. Please refer it.

internal/postgresql/postgresql.go
```diff
@@ -53,6 +53,10 @@ var (
 
 var log = slog.S()
 
+type PGManager interface {
```
sgotti: Can you please add a `go:generate` comment su we know how to generate the mocks? Like in https://github.com/sorintlab/stolon/blob/44f222c92afa0afd7b6cfd13f93e29a5e443619d/internal/store/store.go#L25

dineshba: Added 👍

sgotti:
@dineshba Thanks! Merging.

PR #655, opened 2019-05-31
Update etcd dependency

lawrencejones:
```diff
- github.com/coreos/etcd v3.2.11+incompatible
+ github.com/coreos/etcd v3.3.13+incompatible
```

Between v3.2.11 and v3.3.13 there have been a number of changes to the
way the etcd client handles failed server connections, notably a
significant refactor of the client balancer which appear to have fixed
several issues when communicating with etcd through a TCP L4 load
balancer.


lawrencejones:
@sgotti we were having real issues with our stolon-proxy instances that are resolved with this update.

Our proxies speak to an etcd cluster inside GKE via GCP's TCP L4 load balancer. When performing deployments or in brief moments of network flakiness, the stolon-proxy's etcd client would mark the endpoint as unhealthy and not pin an alternative endpoint, leaving the proxy to fail a clusterdata get twice and fence our connections (ouch).

I'm pretty sure this fixes the issue but I'm going to run some extensive tests overnight to confirm it. Either way, the changes in v3.3 are mostly about client robustness and tolerance of network partitions, something stolon will benefit greatly from.

lawrencejones:
Slightly disappointing report here: this has slightly improved our situation, in that the etcd balancer appears to be making more sensible decisions about what connection to use, but we're still seeing issues via the L4 load balancer.

Still recommend we merge this though, especially given the improvements related to network partition tolerance.

sgotti:
@lawrencejones Thanks for you PR! I'm ok for updating the etcd dependencies.

Related to your issues we don't use a loadbalancer to connect to etcd endpoints but just use the etcd client providing all the endpoints as its loadbalancing logic detects dead members and fastly switches to another endpoint/new etcd leader and never gave us issues.

lawrencejones:
> we don't use a loadbalancer to connect to etcd endpoints

This would be ideal, but we're running our etcd cluster inside a GCP GKE cluster without native VPC (where the pod IP addresses are made available on the GCP project network, and contactable from outside the kubernetes cluster). To provide a pathway into the cluster we have to expose the kubernetes service via a GCP L4 load balancer and that's been the root of some of our issues :(

> as its loadbalancing logic detects dead members and fastly switches to another endpoint/new etcd leader

Agreed, from what I can see the etcd client is very good at this. The updates in this version of etcd have improved that logic though, for example no longer un-pinning bad endpoints for 3s before placing them back in the pool. I was hoping it might fix our problems but it's a good idea to upgrade regardless :)

sgotti:
@lawrencejones Just a small nit. I tried your pr and just noticed that when running 

`go get github.com/coreos/etcd@v3.3.13` 

I got this error:

```
go: finding github.com/coreos/etcd v3.3.13                     
# github.com/coreos/etcd/mvcc/backend                                                  
../../../../go/pkg/mod/github.com/coreos/etcd@v3.3.13+incompatible/mvcc/backend/config_linux.go:31:2: unknown field 'NoFreelistSync' in struct literal of type bolt.Options
```

Build happens anyway since we are only compiling the client.

To fix this also `github.com/coreos/bbolt` must be updated. The unique final diff from your PR is this. The vendor dirs are already identical since go mod noticed that we are not using bbolt and removed it from vendoring.

```diff
diff --git a/go.mod b/go.mod
index 8066136..75a3275 100644
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
        github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da // indirect
        github.com/boltdb/bolt v1.3.1 // indirect
        github.com/cockroachdb/cmux v0.0.0-20170110192607-30d10be49292 // indirect
-       github.com/coreos/bbolt v1.3.0 // indirect
+       github.com/coreos/bbolt v1.3.2 // indirect
        github.com/coreos/etcd v3.3.13+incompatible
        github.com/coreos/go-semver v0.0.0-20150725033620-d043ae190b32 // indirect
        github.com/coreos/go-systemd v0.0.0-20181004131557-b14d3eb023cc // indirect
diff --git a/go.sum b/go.sum
index 1d1cc7b..bea3b20 100644
--- a/go.sum
+++ b/go.sum
@@ -21,6 +21,8 @@ github.com/cockroachdb/cmux v0.0.0-20170110192607-30d10be49292 h1:dzj1/xcivGjNPw
 github.com/cockroachdb/cmux v0.0.0-20170110192607-30d10be49292/go.mod h1:qRiX68mZX1lGBkTWyp3CLcenw9I94W2dLeRvMzcn9N4=
 github.com/coreos/bbolt v1.3.0 h1:HIgH5xUWXT914HCI671AxuTTqjj64UOFr7pHn48LUTI=
 github.com/coreos/bbolt v1.3.0/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
+github.com/coreos/bbolt v1.3.2 h1:wZwiHHUieZCquLkDL0B8UhzreNWsPHooDAG3q34zk0s=
+github.com/coreos/bbolt v1.3.2/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
 github.com/coreos/etcd v3.2.11+incompatible h1:S3HFWk+1nkDWvYPJnWNNqi/dWnXJa3S3zQgFOycHbC8=
 github.com/coreos/etcd v3.2.11+incompatible/go.mod h1:uF7uidLiAD3TWHmW31ZFd/JWoc32PjwdhPthX9715RE=
 github.com/coreos/etcd v3.3.13+incompatible h1:8F3hqu9fGYLBifCmRCJsicFqDx/D68Rt3q1JMazcgBQ=

```

Starting from scratch these are the command I had to do:

```
go get github.com/json-iterator/go
go get github.com/coreos/bbolt
go get github.com/coreos/etcd@v3.3.13
```

and then, since we are using the vendor directory (but I'm going to remove it in the future) a `go mod vendor`

lawrencejones:
Thanks for catching! I've pushed the change now 🙌 

Also, have you considered using https://dependabot.com/ to manage dependency updates for stolon? It's a great tool for automating these sort of changes and ensuring you stay on top of security updates. Here's an example of it doing the job I failed to do in stolon-pgbouncer https://github.com/gocardless/stolon-pgbouncer/pull/50

sgotti:
@lawrencejones Thanks! merging.

> Also, have you considered using https://dependabot.com/ to manage dependency updates for stolon? It's a great tool for automating these sort of changes and ensuring you stay on top of security updates. Here's an example of it doing the job I failed to do in stolon-pgbouncer [gocardless/stolon-pgbouncer#50](https://github.com/gocardless/stolon-pgbouncer/pull/50)

Yeah I though about these tools (also renovate-bot). It's just that I'm not sure it will cope with also the `go mod vendor` we do (perhaps it's time to remove vendoring...) or with the dependency hell that the k8s go client creates (that will be improved now that go modules are used).

PR #656, opened 2019-05-31
sentinel: Add Prometheus metrics

benwh:
These metrics provide the ability to build alerts that tell us whether the sentinels are operating as expected:
- The last time that the sentinel successfully processed the clusterdata.
- Whether the sentinel is a leader.
- The number of times that the sentinel has been elected leader.

---

This follows on from the keeper metrics in #639.

We've found these sentinel metrics to be extremely useful, upon discovering that our sentinels could occasionally all become stuck - and no decisions would be made across the cluster - when there are issues communicating with etcd (we're planning to upstream a patch for that issue in the near future!)

### DIFF
--- a/cmd/sentinel/cmd/metrics.go
+++ b/cmd/sentinel/cmd/metrics.go
@@ -1,0 +1,73 @@
+// Copyright 2019 Sorint.lab
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied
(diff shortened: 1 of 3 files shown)


benwh:
This is what the metrics look like in action, with 3 sentinels running:

Screenshot 2019-05-31 17 10 56

cmd/sentinel/cmd/metrics.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2017 Sorint.lab
```
sgotti: s/2017/2019

benwh: Good spot - fixed!

sgotti:
@benwh Can you please squash in a single commit?

benwh:
@sgotti I'd originally kept it separate as it was also changing a file outside of the scope of this PR. But happy to do so, squashed now.

sgotti:
@benwh Oh I haven't noticed that you also changed the keeper metrics file. Anyway it's not a big styling issue. I'm going to merge it. Thanks again!

benwh:
Excellent, thanks very much for the speedy review!

PR #657, opened 2019-05-31
Small dead code removal in Sentinel.go

dineshba:
Hi @sgotti,
I was going through the code base to better understand the stolon. I came across `one dead code` in **Sentinel.go**. 

To verify that 
- First, I wrote unit tests for that method (first commit in this PR)
- Second, I removed that dead code and checked. All the tests are passing (second commit)

Please lemme know if I missed any test cases.

### DIFF
--- a/cmd/sentinel/cmd/sentinel.go
+++ b/cmd/sentinel/cmd/sentinel.go
@@ -327,7 +327,6 @@ func (s *Sentinel) updateKeepersStatus(cd *cluster.ClusterData, keepersInfo clus
 func (s *Sentinel) activeProxiesInfos(proxiesInfo cluster.ProxiesInfo) cluster.ProxiesInfo {
 	pihs := s.proxyInfoHistories.DeepCopy()
 
-	tmpPihs := pihs.DeepCopy()
 	// remove missing proxyInfos from the history
 	for proxyUID, _ := range pihs {
 		if _, ok := proxiesInfo[proxyUID]; !ok {
@@ -335,7 +334,6 @@ func (s *Sentinel) activeProxiesInfos(proxiesInfo cluster.ProxiesInfo) cluster.P
 		}
 
 	}
-	pihs = tmpPihs
 
 	activeProxiesInfo := proxiesInfo.DeepCopy()
 	// keep only updated proxies info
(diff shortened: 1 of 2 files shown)


sgotti:
@dineshba What that piece of code does is to remove proxyinfos  that are not in the store anymore from the local sentinel `s.proxyInfoHistories` (in etcd/consul the proxyinfos are saved with a ttl so if the proxies stops their entries will expire, in k8s when the pod is deleted the proxyinfo will disappear since they are saved in the pod metadata).

To implement a test for this you should do a two step call to `s.activeProxiesInfos` since this is a stateful function that reads/writes to `s.proxyInfoHistories`. In the first call you could populate with 2 proxies, in the second call just provide 1 of the 2 proxies then `s.proxyInfoHistories` should contain only the remaining proxy.

IMHO this functions can be confusing because it does two things at a time (update `s.proxyInfoHistories` and populate the returned `activeProxiesInfo`)

dineshba:
Cool @sgotti I got the scenario of `removing from sentinel's history if the proxy is removed in store`.

Please refer my fifth test `should remove proxy from sentinel's local history if the proxy is removed in store` https://github.com/sorintlab/stolon/pull/657/files#diff-07b1814671e21d1e829af15ef417877dR5043

Sentinel is given with 2 proxies (initial state of sentinel)(proxy1, proxy2), then called `activeProxiesInfos` with only one proxy (proxy2), then we are expecting `activeProxies` and `proxyInfoHistories` to have only `proxy2`. For this expectation 5th test failed. I fixed it by the removing this [line](https://github.com/sorintlab/stolon/pull/657/files#diff-9961d4546ef09e0152c558e820699523L338).

Is my understanding correct? Please correct me if I am wrong.

sgotti:
@dineshba Uh! That code is definitely wrong since it changes `pihs` and then replaces it with `tmpPihs` that contains the previous value... So I suspect that `s.proxyInfoHistories` currently continues to grow without removing anything...

Thanks for the catch. If you're ok I'll merge your PR.

dineshba:
We can merge this PR @sgotti

sgotti:
@dineshba Thanks again! Merging.

PR #658, opened 2019-05-31
Allow special characters in pg-su-username

dineshba:
If we are having special characters in su username like below
```bash
./stolon-keeper --pg-su-username=user-name # other flags
```
We are getting below error
```
INFO    postgresql/postgresql.go:534    setting superuser password
2019-06-01 01:34:40.031 IST [60923] ERROR:  syntax error at or near "-" at character 14
```

But in postgres, having role as `user-name` is allowed. So it is better if stolon also allow this.

I just added double quotes for `username` in `alter role` command.

> There are two more places in `stolon` where we are altering role. In those two places we are already using double quotes. https://github.com/sorintlab/stolon/blob/master/internal/postgresql/utils.go#L104 and https://github.com/sorintlab/stolon/blob/master/internal/postgresql/utils.go#L115

### DIFF
--- a/internal/postgresql/utils.go
+++ b/internal/postgresql/utils.go
@@ -68,7 +68,7 @@ func setPassword(ctx context.Context, connParams ConnParams, username, password 
 	}
 	defer db.Close()
 
-	_, err = dbExec(ctx, db, fmt.Sprintf(`alter role %s with password '%s';`, username, password))
+	_, err = dbExec(ctx, db, fmt.Sprintf(`alter role "%s" with password '%s';`, username, password))
 	return err
 }
 


sgotti:
@dineshba Good catch!

What about using consistent single quotes (or double quotes, I don't know which is better in such case) like in `with password '%s'`.

As a note if someone is going to ask:

I'll have preferred to use sql prepared statements with arguments like: 
`dbExec(ctx, db, "alter role $1 with password $2", username, password)`

but unfortunately this doesn't work (returning a syntax error) since in postgres prepared statements works only for SELECT, INSERT, UPDATE, DELETE, or VALUES statement: https://www.postgresql.org/docs/11/sql-prepare.html

dineshba:
Thanks for the quick review @sgotti. For consistency, I made everything as `single quotes` and amended to the same commit.

sgotti:
@dineshba you're right. Merging. Thanks!

dineshba:
@sgotti Thanks 👍

PR #664, opened 2019-06-05
keeper: configurable wal_keep_segments

sgotti:
use the value provided in the pg parameters to set the instance
wal_keep_segments value.

The value is ignored if it's a bad value or lesser than the default value (8).

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -57,7 +57,10 @@ var CmdKeeper = &cobra.Command{
 	Version: cmd.Version,
 }
 
-const maxPostgresTimelinesHistory = 2
+const (
+	maxPostgresTimelinesHistory = 2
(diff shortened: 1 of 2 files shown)


rearden-steel:
@sgotti 
Thanks, we need this PR very much :)

cmd/keeper/cmd/keeper.go
```diff
@@ -57,7 +57,10 @@ var CmdKeeper = &cobra.Command{
 	Version: cmd.Version,
 }
 
-const maxPostgresTimelinesHistory = 2
+const (
+	maxPostgresTimelinesHistory = 2
+	defaultWalKeepSegments      = 8
```
lawrencejones: @hmac had a suggestion that this should maybe be `minWalKeepSegments` now?

sgotti: @lawrencejones Updated.

PR #665, opened 2019-06-05
Enable verbose & progress logging for pg_basebackup

hmac:
These two changes cause `pg_basebackup` to log the names of the files it is processing, along with the estimated progress through the entire backup. For large databases this can be very useful in estimating how long the backup will take.

### DIFF
--- a/internal/postgresql/postgresql.go
+++ b/internal/postgresql/postgresql.go
@@ -829,7 +829,7 @@ func (p *Manager) SyncFromFollowed(followedConnParams ConnParams, replSlot strin
 
 	log.Infow("running pg_basebackup")
 	name := filepath.Join(p.pgBinPath, "pg_basebackup")
-	args := []string{"-R", "-Xs", "-D", p.dataDir, "-d", followedConnString}
+	args := []string{"-R", "-v", "-P", "-Xs", "-D", p.dataDir, "-d", followedConnString}
 	if replSlot != "" {
 		args = append(args, "--slot", replSlot)
 	}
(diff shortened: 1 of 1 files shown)


sgotti:
@hmac Thanks for your PR. I think it's a good idea to show progress.

The unique small issue it that progress reporting on a tty uses line feed control char to update its progress state (rewrite the current line without carriage return) so it'll probably mess up/remove some concurrent logging output from the keeper. This doesn't happen when not on a tty (pg_basebackup detects if stdout is a tty or not).

One possible solution could be to set pg_basebackup stdout/stderr to the same fd of the keeper but pipe them to a goroutine that will log them.

internal/postgresql/postgresql.go
```diff
@@ -838,10 +838,24 @@ func (p *Manager) SyncFromFollowed(followedConnParams ConnParams, replSlot strin
...
+	// itself (in some modes; we don't use this). As a result we only need to deal with
+	// stderr.
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("error: %v", err)
+	}
+
+	cmd.Start()
+	if _, err := io.Copy(os.Stderr, stderr); err != nil {
+		return fmt.Errorf("error: %v", err)
```
sgotti: Returning an error here (also it's vary hard that this will happen) will keep a zombie process since `cmd.Wait()` will not be called.

I'll prefer to execute the io.Copy inside a go routine so the program function can go ahead and block on `cmd.Wait()`. The goroutine will ignore possible io.Copy errors since they're not really important.

hmac:
@sgotti thanks for the review and apologies for the sloppy code! I'm still getting used to Go 😅 

What I've done is pass a context to the `exec.Command` and cancel it when we return from the function, to ensure that it gets cleaned up. We should then block on `io.Copy` until the process completes, but we'll be streaming to stderr the whole time so we'll still get incremental progress logs.

I think this covers all your points, but please let me know if I've missed anything!

sgotti:
> What I've done is pass a context to the `exec.Command` and cancel it when we return from the function, to ensure that it gets cleaned up. We should then block on `io.Copy` until the process completes, but we'll be streaming to stderr the whole time so we'll still get incremental progress logs.

passing a context to exec.Command is used to kill the command when the context expires but it won't implicitly wait for it so it'll leave a zombie process anyway. So using a context are is not useful and confusing.

If you want to avoid adding a goroutine then a solution is to do the io.Copy like you're doing now. io.Copy will return without an error when the process exits and then the function will pass to the cmd.Wait. If io.Copy returns an error then just log it without returning so cmd.Wait() will be called anyway.

hmac:
> passing a context to exec.Command is used to kill the command when the context expires but it won't implicitly wait for it so it'll leave a zombie process anyway. So using a context are is not useful and confusing.

sgotti:
> Just for my own understanding, do you mean that the process will be killed, but this might take some time and meanwhile stolon would have carried on?

At the lower level killing a process is a kill syscall that won't block so the function could exit while the process is still alive. That's why you must always call cmd.Wait to be sure that the process is exited, be able to get its exit code and avoid having a zombie child process.

But what I was trying to says is that the unique purpose of attaching a context to a cmd is to have the process killed when it expires. But that's not our need so a context is not needed.

hmac:
@sgotti thank you for your comments so far and sorry this PR has taken so long! I've update the changes to match what I think you intended - does it look like the right approach?

internal/postgresql/postgresql.go
```diff
@@ -838,11 +838,25 @@ func (p *Manager) SyncFromFollowed(followedConnParams ConnParams, replSlot strin
...
+	// stderr.
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return err
+	}
+
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	go io.Copy(os.Stderr, stderr)
```
sgotti: I'll also log the io.Copy error. Just in case.

lawrencejones:
`io.Copy` should now be logged @sgotti. Hope this now looks good?

sgotti:
@hmac @lawrencejones Thanks! Merging

PR #666, opened 2019-06-06
Add Timeout for kubeclient

dineshba:
Adding timeout for `kubeclient` by adding timeout for `restclient`

I have made two commits
- Move common logic into a method called `getKubeValues`
- Added timeout for kubecfg

> Note: I have tested the kubecfg by providing different timeout value and it works.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -166,23 +166,7 @@ func NewStore(cfg *CommonConfig) (store.Store, error) {
 		}
 		s = store.NewKVBackedStore(kvstore, storePath)
 	case "kubernetes":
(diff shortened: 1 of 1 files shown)


cmd/common.go
```diff
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/sorintlab/stolon/internal/cluster"
```
sgotti: please reorganize imports to split between stdlib, stolon packages and external packages.

dineshba: Sure. I missed it :-( . Thanks for the catch. Updated and amended the commit

PR #667, opened 2019-06-06
Release v0.14.0

sgotti:
WIP changelog for release v0.14.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,52 @@
 ### v0.14.0
 
+#### New features
+
+* Add sentinel prometheus metrics [#656](https://github.com/sorintlab/stolon/pull/656))
+* Store only the last 2 postgres timeline histories to not exceed the max value size of the underlying store [#651](https://github.com/sorintlab/stolon/pull/651))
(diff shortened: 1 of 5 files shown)


maksm90:
Hi @sgotti!
Unfortunately, the autorestart feature #568 is not safe for parameters (`max_connections`, `max_prepared_transaction`, etc described [here](https://www.postgresql.org/docs/current/hot-standby.html#HOT-STANDBY-ADMIN)) that require the special order of restart of nodes (master -> replicas or back). This feature requires some further development.
I propose to mention explicitly in doc to not use this feature for those problem postgres parameters until it will be done correctly.

arunvelsriram:
@sgotti just wanted to bring it to your notice that #651 was co authorthored.

daMupfel:
Is there any info about when this sweet release might be published :smile:  ?

sgotti:
@maksm90 Thanks. Can you please open a new issue so we can track it? (I was thinking that the next restart after a failed one will work)

@arunvelsriram Thanks. I'll add you.

@daMupfel Probably next week.

PR #669, opened 2019-06-10
Label cluster identifier with component

lawrencejones:
[^1]: https://github.com/gocardless/stolon-pgbouncer/pull/46

The stolon_cluster_identifier metric is universal to all stolon
components, and can be used to join metric series across a stolon
cluster. By adding the component label, it's possible to write
polymorphic alerts that expose the component name when firing.

This change also registers the metric in the sentinel and proxy, as it
should be present in all long-lived binaries. This accompanies a change
in stolon-pgbouncer to do the same[^1].


sgotti:
@lawrencejones LGTM. Thanks. Merging.

PR #675, opened 2019-06-21
Option on cluster spec to configure proxy check interval and timeout

trmfreitas:
Hi!

I added configuration of proxy request timeout and proxy check interval set on Cluster spec.

Configured as option on clusterspec:
- proxyTimeoutInterval
- proxyCheckInterval

Also changed documentation.

This should fix:
#314

And help on:
#588

Thanks.

Best regards,
Tiago Freitas


chenguang-Steve:
Hi @trmfreitas 
I use kubernetes store backend. Seems that it is not stable and encounter proxy check timeout many times. It will cause reset database connections making program fail. So it is best to apply your change.
But from proxy logs, it seems timeout after 5 seconds instead default 15s. Don't know why.

`2020-01-10T05:17:13.932Z	INFO	cmd/proxy.go:236	master address	{"address": "10.244.63.219:5432"}
2020-01-10T05:17:19.168Z	INFO	cmd/proxy.go:264	check timeout timer fired
2020-01-10T05:17:19.475Z	INFO	cmd/proxy.go:151	Stopping listening
2020-01-10T05:17:20.096Z	INFO	cmd/proxy.go:303	check function error	{"error": "failed to update proxyInfo: update failed: failed to get latest version of pod: Get https://10.96.0.1:443/api/v1/namespaces/collateral/pods/collateral-postgres-postgresql-proxy-77d5cc5d97-pzvnf: net/http: request canceled (Client.Timeout exceeded while awaiting headers)"}
2020-01-10T05:17:31.046Z	INFO	cmd/proxy.go:303	check function error	{"error": "cannot get cluster data: failed to get latest version of configmap: Get https://10.96.0.1:443/api/v1/namespaces/collateral/configmaps/stolon-cluster-collateral-postgres-postgresql: net/http: request canceled (Client.Timeout exceeded while awaiting headers)"}
2020-01-10T05:17:36.783Z	I`

sgotti:
@trmfreitas Sorry I lost this PR. I haven't reviewed it yet, can you please rebase it?

@chenguang-Steve If your k8s api are unreliable/slow you should improve them or just use a dedicated store like etcd. See the architecture doc note https://github.com/sorintlab/stolon/blob/master/doc/architecture.md#kubernetes-store-backend:

`
**NOTE**: a kubernetes based store will share the kubernetes API server with all the other k8s cluster resources. If the API servers are overloaded and doesn't answer in time, as explained above, the proxies will, after a timeout, close connections to the master keeper. The same will happen if you're updating you kubernetes cluster and you have the need to shutdown all your API servers. A dedicated store, like an etcd cluster, will probably be more reliable and avoid the proxies closing connection and impacting you application availability.
`

snailxr:
@trmfreitas We need these config,can you review the pull request 😀

cmd/proxy/cmd/proxy.go
```diff
@@ -89,6 +89,9 @@ type ClusterChecker struct {
 	endPollonProxyCh chan error
 
 	pollonMutex sync.Mutex
+
+	checkIntervalSeconds  int
+	requestTimeoutSeconds int
```
sgotti: Just use time.Duration instead of an int. So you can avoid all the conversions done in the other parts of the code you changed.

sgotti: Please call them as the config options: ProxyCheckInterval and ProxyTimeoutInterval.

cmd/proxy/cmd/proxy.go
```diff
@@ -305,7 +319,17 @@ func (c *ClusterChecker) Start() error {
...
-			timerCh = time.NewTimer(cluster.DefaultProxyCheckInterval).C
+
+			if previousCheckIntervalSeconds != c.checkIntervalSeconds {
+				log.Infow("updated check interval seconds is ", "seconds", c.checkIntervalSeconds)
+				previousCheckIntervalSeconds = c.checkIntervalSeconds
+			}
+			if previousRequestTimeoutSeconds != c.requestTimeoutSeconds {
+				log.Infow("updated request timeout seconds is ", "seconds", c.requestTimeoutSeconds)
+				previousRequestTimeoutSeconds = c.requestTimeoutSeconds
+			}
```
sgotti: Make these logs at Debug level

cmd/proxy/cmd/proxy.go
```diff
@@ -285,6 +293,12 @@ func (c *ClusterChecker) Start() error {
 	checkCh := make(chan error)
 	timerCh := time.NewTimer(0).C
 
+	previousCheckIntervalSeconds := c.checkIntervalSeconds
+	previousRequestTimeoutSeconds := c.requestTimeoutSeconds
+
+	log.Infow("check interval seconds is ", "seconds", c.checkIntervalSeconds)
+	log.Infow("request timeout seconds is ", "seconds", c.requestTimeoutSeconds)
```
sgotti: Make these logs at Debug level

doc/cluster_spec.md
```diff
@@ -15,7 +15,9 @@ Some options in a running cluster specification can be changed to update the des
...
-| requestTimeout            | time after which any request (keepers checks from sentinel etc...) will fail.                                                                                                                                                                                                                                                                                                                                                                                                     | no                        | string (duration) | 10s                                                                                                                                 |
+| requestTimeout            | time after which any request (keepers checks from sentinel etc...) will fail. | no                        | string (duration) | 10s                     | 
+| proxyTimeoutInterval      | time after which a proxy will fail when communicating with database | no                        | string (duration) | 15s                               |
```
sgotti: This isn't the time the proxy fails communicating with the database but the maximum interval where the proxy should complete a cycle of reading from the store and update its state to the store.

doc/cluster_spec.md
```diff
@@ -15,7 +15,9 @@ Some options in a running cluster specification can be changed to update the des
...
+| proxyCheckInterval        | proxy check interval | no                        | string (duration) | 5s                               |
```
sgotti: This is the interval where the proxy will check for updates from the store. It must be smaller then proxyTimeoutInterval.

cmd/proxy/cmd/proxy.go
sgotti: I'll also log a warning if proxyCheckInterval isn't smaller than proxyTimeoutInterval since the proxy will always fail.

internal/cluster/cluster.go
```diff
@@ -215,6 +215,10 @@ type ClusterSpec struct {
 	SleepInterval *Duration `json:"sleepInterval,omitempty"`
 	// Time after which any request (keepers checks from sentinel etc...) will fail.
 	RequestTimeout *Duration `json:"requestTimeout,omitempty"`
+	// Proxy check interval
+	ProxyCheckInterval *Duration `json:"proxyCheckInterval,omitempty"`
+	// Proxy check timeout interval
```
sgotti: Proxy timeout interval

sgotti:
@trmfreitas Since this feature requires more changes to the logic and interactions between the proxy and the sentinel that are quite difficult to describe to someone else that doesn't know how stolon internally works I preferred to handle it by myself and created PR #756. Thanks for your effort.

PR #677, opened 2019-06-23
use abronan/valkeyrie instead of docker/libkv and add support redis

vahid-sohrabloo:


sgotti:
@vahid-sohrabloo Thanks for your PR. I'll be on vacation until next week so I won't be very responsive. I saw this fork of libkv but ignored it since it doesn't provide a context interface. So I won't use it to replace etcd v3. It could replace libkv for current libkv stores if nothing better (using context.Context) exists and it provides some bug fixes.

For adding redis as a store I'll do it in another PR since it'll require also all the integration tests setup.

vahid-sohrabloo:
@sgotti  thanks for quick response. I work on Valkyria to add context. I'll be creating PR in a few days

PR #681, opened 2019-07-05
Publish proxy addresses in the store's clusterdata.

arssher:
This is useful for applications willing to learn cluster entrypoint address.

To this end, dummy ProxyStatus struct in cd.Proxy is turned into map proxyUid ->
ProxyStatus. And since cd now contains data from ProxyInfo, passing active
proxies from activeProxiesInfos to updateCluster separately is not needed
anymore.

### DIFF
--- a/cmd/proxy/cmd/proxy.go
+++ b/cmd/proxy/cmd/proxy.go
@@ -48,10 +48,11 @@ var CmdProxy = &cobra.Command{
 type config struct {
 	cmd.CommonConfig
 
-	listenAddress string
-	port          string
-	stopListening bool
-	debug         bool
+	listenAddress    string
+	advertiseAddress string
+	port             string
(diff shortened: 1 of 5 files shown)


sgotti:
@arssher Thanks for the PR. But I don't see a reason why the proxy listen address should be published to the cluster data. It's not really needed anywhere for stolon. You should publish it using other ways (ie a k8s service) but the stolon cluster data is not the right place.

arssher:
Well, this is indeed not needed for Stolon itself, however since Stolon currently doesn't expose any kind of special API for getting current cluster state (who is alive, who is the master, etc), cluster data is read by external tools, e.g. monitoring. And it is very convenient to be able to learn where is the cluster entrypoint from there -- see it like service discovery.

PR #689, opened 2019-08-11
add DisableAlterSystemSet cluster_spec option

idetoile:
add `DisableAlterSystemSet` to cluster_spec options
related to https://github.com/sorintlab/stolon/issues/683

Issue #683, opened 2019-07-11
"CREATE EXTENSION citus;" fails


yuzhichang:
### Submission type
 - Bug report

### Environment
Postgresql 11 cluster with stolon v0.13.0.

### Expected behaviour you didn't see
`CREATE EXTENSION citus;` succeed

### Unexpected behaviour you saw
```
postgres# CREATE EXTENSION citus;
ERROR:  could not fsync file "postgresql.auto.conf": Invalid argument
CONTEXT:  SQL statement "SELECT citus_setup_ssl()"
PL/pgSQL function inline_code_block line 5 at PERFORM
```
The citus extension sql `citus--8.0-9.sql` calls `AlterSystemSetConfigFile` (defined in postgresql source file `src/backend/utils/misc/guc.c`) to enable SSL. `AlterSystemSetConfigFile` modifies postgresql.auto.conf.

sgotti:
@yuzhichang stolon disables writing to postgresql.auto.conf since an `alter system set` will affect only the instance where it's executed. Instead users should use the cluster spec pgParameters options so all the instances will use the same parameters.

You have two options:

* find a way to stop the extension on executing the command and set the required parameters in the clusterspec
* modify the stolon code to not do this (or better create a pull request to add an option to the clusterspec to choose the behavior).

idetoile:
A simple workaround ;), to run on keepers:
```sh
rm $PGDATA/postgres/postgresql.auto.conf
```

idetoile:
I've just made a pull request for this feature

Another solution (temp solution) is using this fork:
https://github.com/infinivision/stolon
https://github.com/infinivision/stolon/releases

Issue #690, opened 2019-08-13
cannot get configured pg parameters


idetoile:
### Submission type

 - Bug report

#### Stolon version

master branch fork (after v0.14.0)

#### Additional environment information if useful to understand the bug

semaphoreci

### Expected behaviour you didn't see

consul test passing

### Unexpected behaviour you saw

failed cause of
`cmd/keeper.go:673	cannot get configured pg parameters	{"error": "dial unix /tmp/.s.PGSQL.2343: connect: no such file or directory"}`

### Steps to reproduce the problem

run integration test on semaphoreci many times, because it's happening randomly

you can see it here https://semaphoreci.com/sorintlab/stolon/branches/pull-request-689/builds/10

related to https://github.com/sorintlab/stolon/pull/689#issuecomment-520750565

idetoile:
Can't reproduce, I think it was my bad

PR #693, opened 2019-08-24
STKEEPER_DISABLE_ALTER_SYSTEM option

ghost:
STKEEPER_DISABLE_ALTER_SYSTEM option
final and working solution for issue https://github.com/sorintlab/stolon/issues/683
and following wrong approach in pull https://github.com/sorintlab/stolon/pull/689

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -98,7 +98,7 @@ type config struct {
 
 	uid                     string
 	dataDir                 string
-	disableAlterSystem      bool
+	enableAlterSystem       bool
 	debug                   bool
 	pgListenAddress         string
 	pgAdvertiseAddress      string
@@ -143,7 +143,7 @@ func init() {
 	CmdKeeper.PersistentFlags().StringVar(&cfg.pgSUUsername, "pg-su-username", user, "postgres superuser user name. Used for keeper managed instance access and pg_rewind based synchronization. It'll be created on db initialization. Defaults to the name of the effective user running stolon-keeper. Must be the same for all keepers.")
 	CmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPassword, "pg-su-password", "", "postgres superuser password. Only one of --pg-su-password or --pg-su-passwordfile must be provided. Must be the same for all keepers.")
 	CmdKeeper.PersistentFlags().StringVar(&cfg.pgSUPasswordFile, "pg-su-passwordfile", "", "postgres superuser password file. Only one of --pg-su-password or --pg-su-passwordfile must be provided. Must be the same for all keepers)")
-	CmdKeeper.PersistentFlags().BoolVar(&cfg.disableAlterSystem, "disable-alter-system", true, "disable ALTER SYSTEM command. Since postgresql.auto.conf overrides postgresql.conf parameters, changing some of them with ALTER SYSTEM could break the cluster and make pg parameters different between the instances. (use it with caution !!!)")
+	CmdKeeper.PersistentFlags().BoolVar(&cfg.enableAlterSystem, "enable-alter-system", false, "enable ALTER SYSTEM command. Since postgresql.auto.conf overrides postgresql.conf parameters, changing some of them with ALTER SYSTEM could break the cluster and make pg parameters different between the instances. (use it with caution !!!)")
 	CmdKeeper.PersistentFlags().BoolVar(&cfg.debug, "debug", false, "enable debug logging")
 
 	CmdKeeper.PersistentFlags().MarkDeprecated("id", "please use --uid")
@@ -445,7 +445,7 @@ type PostgresKeeper struct {
 	sleepInterval  time.Duration
 	requestTimeout time.Duration
 
-	disableAlterSystem bool
+	enableAlterSystem bool
 
 	e   store.Store
 	pgm *postgresql.Manager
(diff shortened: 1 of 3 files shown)


internal/postgresql/postgresql.go
```diff
@@ -61,6 +61,7 @@ type PGManager interface {
 
 type Manager struct {
 	pgBinPath             string
+	disableAlterSystem    bool
```
lawrencejones: Can we pull this line to the bottom of the struct fields please, under requestTimeout? The fields are vaguely ordered in importance, which is why dataDir and parameters etc are up top. This type of flag is really optional, and is best placed as something developers see last.

cmd/keeper/cmd/keeper.go
```diff
@@ -477,6 +481,8 @@ func NewPostgresKeeper(cfg *config, end chan error) (*PostgresKeeper, error) {
 
 		dataDir: dataDir,
 
+		disableAlterSystem: cfg.disableAlterSystem,
```
lawrencejones: Pull this line down to just below the `requestTimeout` please!

cmd/keeper/cmd/keeper.go
```diff
@@ -424,6 +426,8 @@ type PostgresKeeper struct {
 
 	bootUUID string
 
+	disableAlterSystem bool
```
lawrencejones: Below requestTimeout please

internal/postgresql/postgresql.go
```diff
@@ -306,8 +308,10 @@ func (p *Manager) start(args ...string) error {
 	// the instance parent is the keeper instead of the defined system reaper
 	// (since pg_ctl forks and then exits leaving the postmaster orphaned).
 
-	if err := p.createPostgresqlAutoConf(); err != nil {
-		return err
+	if p.disableAlterSystem {
```
lawrencejones: We should add some code to manage the removal of the symlink if someone was previously running with `disable-alter-system=true`, and now wants to run with it enabled.

In the current form, if Stolon users were to upgrade to a Stolon version with this PR, they'd have lots of `/dev/null` symlink files remaining on their machines. I think it's best if we can detect when the symlink to `/dev/null` is there and replace it with a blank `postgresql.auto.conf` file when we've got alter system enabled, as this would be least surprising to me as a user.

lawrencejones:
Fwiw, wanting to run alter system commands is a use case we've encountered a few times now. I can't speak to the citrus extension, but we occasionally want to verify the effect of a change on just the primary/replica first before rolling it out across the cluster.

You make a good point though- the average stolon user probably shouldn't enable alter system for the reasons you mention. We should adapt the docs to make it clear this isn't recommended.

cmd/keeper/cmd/keeper.go
idetoile: ok, done just right now

cmd/keeper/cmd/keeper.go
idetoile: ok, done just right now

internal/postgresql/postgresql.go
idetoile: ok, added this behaviour just right now

idetoile:
@lawrencejones thanks for reviewing, the feedback was pretty clear :)
@sgotti thanks for feedback, I've just added more documentation to the option from #343, you're right, I've just tested now and saw that citus doesn't run alter system if not needed, I wasn't able to make citus work at first time because it enable ssl but it also create certificates automatically, so adding pgParameters at startup (before running the create extension citus command) was failing (no certificate). But I had to make certificate myself and add pgParameters at startup to make restart working. So, finally this option was not required, or maybe to test or see the result of create extension citus in a first time. So I'm changing doc/postgres_parameters.md according to @lawrencejones comment on this feature utility

idetoile:
semaphoreci ununderstandable build failed see https://github.com/sorintlab/stolon/issues/690

cmd/keeper/cmd/keeper.go
```diff
@@ -443,6 +445,8 @@ type PostgresKeeper struct {
 	sleepInterval  time.Duration
 	requestTimeout time.Duration
 
+	disableAlterSystem bool
```
lawrencejones: @sgotti pointed out in a separate PR that flags are usually much more intuitive when they're positive, ie. `enable-alter-system`, with a default `false`.

For consistency, could we change disable->enable please?

doc/postgres_parameters.md
```diff
@@ -49,6 +49,8 @@ Since postgresql.auto.conf overrides postgresql.conf parameters, changing some o
 
 To avoid this stolon disables the execution of ALTER SYSTEM commands making postgresql.auto.conf a symlink to /dev/null. When an ALTER SYSTEM command is executed it'll return an error.
 
+ALTER SYSTEM command can be re-enabled by using `--disable-alter-system=false` parameter to `stolon-keeper` or using `STKEEPER_DISABLE_ALTER_SYSTEM=false` environment variable. This can be useful to occasionally verify the effect of a change on just the primary/replica first before rolling it out across the cluster.
```
lawrencejones: > before rolling it out across the cluster, but should be handled with extreme caution.

Shall we add the warning to the docs? Feels like we should finish with a "beware dragons" just to reiterate that this may not be a good idea for everyone.

sgotti:
@idetoile @lawrencejones Using a keeper option will require a keeper restart. Is this ok for you. Another way will be to use a clusterspec options. Alter system will be enabled for all the instances but you'll execute it only on the instance you want and won't require a restart.

idetoile:
This is ok for me.
The first approach I was trying was based clusterspec https://github.com/sorintlab/stolon/pull/689 but that was breaking many unit testings and had introduced strange and incomprehensible bugs. My level in golang (I just begin) was not enough to fix it, so I finally came to keeper approach and thought that was the best one, your feedback indicating me that is not, but for me it's sufficient.

PR #695, opened 2019-08-29
Reusing FindDB method and added test for the same

dineshba:
Tasks:
1) Reuse the method FindDB
2) Add test coverage for method FindDB


sgotti:
@dineshba Thanks for the PR! It LGTM.

PR #698, opened 2019-09-10
Only pg_rewind against master DBs

lawrencejones:
pg_rewind does not support executing against standby nodes. While stolon
doesn't yet support cascading replication, it will never be correct to
run pg_rewind against any node other than the primary.

By making this change now, we future proof stolon and enable any forks
(such as gocardless/stolon) to configure a cascading replication setup
without issue.

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -828,7 +828,7 @@ func (p *PostgresKeeper) Start(ctx context.Context) {
 	}
 }
 
-func (p *PostgresKeeper) resync(db, followedDB *cluster.DB, tryPgrewind bool) error {
+func (p *PostgresKeeper) resync(db, masterDB, followedDB *cluster.DB, tryPgrewind bool) error {
 	pgm := p.pgm
 	replConnParams := p.getReplConnParams(db, followedDB)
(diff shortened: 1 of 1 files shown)


lawrencejones:
@sgotti we hit this problem in our fork, where we have a setup that permits cascading replication. In our setup, we have this:

```
===== Keepers/DB tree =====

stolon_production_0 (master)
└─stolon_production_1 (sync)
  ├─stolon_async_production_0 (async)
  └─stolon_production_2 (async)
```

Which works great, except when either of `stolon_async_production_0` or `stolon_production_2` attempt to resync against `stolon_production_1`, where `pg_rewind` will fail as it cannot create the necessary temporary tables.

I don't think the addition of this logic is overly burdensome for upstream, prepares us for when we do get round to upstreaming a cascading replication setup (sometime soon I hope!) and prevents anyone from hitting this stumbling block in future.

Depending on your feedback we can either leave this closed until we need it, or get this merged in upstream. We'd appreciate it in upstream as it makes it easier to manage the changes, but I'd understand if you prefer to leave this out for now.

sgotti:
@lawrencejones Thanks! LGTM. Is this ready to be merged?

lawrencejones:
Yup! Let's do it.

PR #701, opened 2019-09-18
Additional info initialize cluster data

sihar:
I fail login to postgres. Then I check log pg_sentinel container and find error "no cluster data available".
This step help me initialize data. Hope help anyone with same issue

### DIFF
--- a/examples/swarm/README.md
+++ b/examples/swarm/README.md
@@ -127,6 +127,7 @@ sf79wnygtcmu        pg_proxy            replicated          2/2                 
 tkbnrfdj4axa        pg_sentinel         replicated          2/2                 sorintlab/master-pg10:latest
 ```
 
+
 ### Connect to the db
 
 #### Connect to the proxy service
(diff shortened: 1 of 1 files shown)


dineshba:
Hi @sihar If someone follow this step to [initialize](https://github.com/sorintlab/stolon/blob/master/examples/swarm/README.md#initialize-the-cluster), then we may not need additional info section right ?

Correct me if I am wrong. Thanks

sihar:
> Hi @sihar If someone follow this step to [initialize](https://github.com/sorintlab/stolon/blob/master/examples/swarm/README.md#initialize-the-cluster), then we may not need additional info section right ?
> 
> Correct me if I am wrong. Thanks

yes @dineshba

sgotti:
@sihar As @dineshba said there's already the step to initialize the cluster so I don't see why we should add more info if the user forgot to execute this step.

@sihar If you want to add additional ways to execute stolonctl inside a container like don for the k8s example please open a new PR. Thanks!

PR #708, opened 2019-10-05
Allow cluster to recover during partial storage failure

drdivano:
podManagementPolicy: Parallel prevents the following situation:

1. We run Stolon as a StatefulSet with 2 or more replicas
2. The cluster goes down together with storage system
3. The cluster comes up. The storage system comes up partially, with Persistent Volume 0 still unavailable.
4. Kubernetes fails to re-schedule and start the StatefulSet since storage for replica 0 is still down.

The Parallel policy tells Kubernetes to start other replicas without waiting for replicas with lower ordinal numbers.

### DIFF
--- a/examples/kubernetes/stolon-keeper.yaml
+++ b/examples/kubernetes/stolon-keeper.yaml
@@ -8,6 +8,7 @@ metadata:
 spec:
   serviceName: "stolon-keeper"
   replicas: 2
+  podManagementPolicy: Parallel
   template:
     metadata:
       labels:


sgotti:
@drdivano Thanks for the PR!. Since it's related to just the k8s example can you rename it to `k8s example: use podManagementPolicy: Parallel`

And then in the description describe why it's useful.

drdivano:
> @drdivano Thanks for the PR!. Since it's related to just the k8s example can you rename it to k8s example: use podManagementPolicy: Parallel
> 
> And then in the description describe why it's useful.

@sgotti I renamed the PR and added a description as you asked 13 days ago. Did I misunderstand you?

sgotti:
> @sgotti I renamed the PR and added a description as you asked 13 days ago. Did I misunderstand you?

Sorry, I just lost your push. Can you please rebase in a single commit instead of merging master branch (for a cleaner git history)? Thanks.

Issue #723, opened 2019-10-23
stolonctl: don't show null cluster spec options


viggy28:
### Submission type

 - Bug report

#### Stolon version
v0.13.0 (and the recent commit too)

Unable to remove these parameters `additionalWalSenders`, `additionalMasterReplicationSlots` & `automaticPgRestart` from cluster spec.

### Expected behaviour you didn't see
As per https://github.com/sorintlab/stolon/blob/master/doc/cluster_spec.md, patching cluster spec with null should remove the parameter whereas it didn't happen

### Steps to reproduce the problem
Cluster spec already has default value of 	
"additionalWalSenders": null, "additionalMasterReplicationSlots": null, and "automaticPgRestart": null. 

stolonctl --cluster-name stolon-cluster --store-backend=etcdv3 update --patch '{"automaticPgRestart":null}' doesn't do anything.

It's likely that https://github.com/sorintlab/stolon/blob/14bc5473237ee3a9a361d83967926940a934077b/internal/cluster/cluster.go#L290 `omitempty` is not mentioned in the JSON tag.

sgotti:
@viggy28 I think that what's confusing is that they are printed as null during a `stolonctl spec`. They were really removed (so I'm going to change your title) and there's no bug or issue on this side.

`omitempty` will avoid showing them as null when marshalled to json and print.

sgotti:
Fixed by #736

PR #724, opened 2019-10-24
Fixing the cluster spec output formatting issue

viggy28:
To address #723.

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -290,6 +290,37 @@ type ClusterSpec struct {
 	AutomaticPgRestart *bool `json:"automaticPgRestart"`
 }
 
+type ClusterSpecNew struct {
+	SleepInterval                    *Duration         `json:"sleepInterval,omitempty"`
+	RequestTimeout                   *Duration         `json:"requestTimeout,omitempty"`
+	ConvergenceTimeout               *Duration         `json:"convergenceTimeout,omitempty"`
+	InitTimeout                      *Duration         `json:"initTimeout,omitempty"`
+	SyncTimeout                      *Duration         `json:"syncTimeout,omitempty"`
+	DBWaitReadyTimeout               *Duration         `json:"dbWaitReadyTimeout,omitempty"`
+	FailInterval                     *Duration         `json:"failInterval,omitempty"`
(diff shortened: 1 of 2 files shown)


viggy28:
Hi @sgotti, thanks for looking. Updated it.

sgotti:
@viggy28 Thinking and testing this PR I noticed that it will break the `stolonctl spec --defaults` output hiding some defaults values.

One possible solution will require the creation (in stolonctl_spec.go), of two different cluster spec structs (identical to the original cluster spec, one with all json tags with omitempty and one without omitempty) that could be [converted](https://golang.org/ref/spec#Conversions) from the original clusterspec so their marshalling will work for both cases (this will also help detecting when we forgot to add new fields to all the structs at compilation time).

viggy28:
Thanks @sgotti. Updated the PR. Let me know if that works.

internal/cluster/cluster.go
```diff
@@ -290,6 +290,86 @@ type ClusterSpec struct {
 	AutomaticPgRestart *bool `json:"automaticPgRestart"`
 }
 
+type ClusterSpecNew struct {
```
sgotti: There's the need of two different cluster spec structs identical to the original cluster spec:

* one with all json tags with omitempty for normal output
* one with all json tags without omitempty for --default output

internal/cluster/cluster.go
```diff
@@ -290,6 +290,86 @@ type ClusterSpec struct {
...
+	// Interval to wait before next check
```
sgotti: Just remove all the fields comments.

viggy28: updated.

sgotti:
> One possible solution will require the creation (in stolonctl_spec.go), of two different cluster spec structs (identical to the original cluster spec, one with all json tags with omitempty and one without omitempty) that could be converted from the original clusterspec so their marshalling will work for both cases (this will also help detecting when we forgot to add new fields to all the structs at compilation time).

@viggy28 Sorry if I wasn't clear but I meant to use two new structs inside stolonctl/cmd/spec.go, take a look at this compare:

https://github.com/sorintlab/stolon/compare/master...sgotti:stolonctl_spec_improve_output?expand=1

sgotti:
@viggy28 Thanks for your PR, I created a PR #736 with an updated implementation.

Issue #604, opened 2019-02-04
How to force a non master keeper to resync with master data?


pedromvieira:
I'm using Stolon with Postgres 11 and Kubernetes.
I've created a new empty disk for stolon-data (with less storage but bigger than database size).
The container / pod starts but dosent sync data (around 13 GB).
How to force this keeper to get master data and sync again?

root@stolon-keeper-0:/stolon-data# stolonctl status --cluster-name kube-stolon --store-backend kubernetes --kube-resource-kind=configmap init
=== Active sentinels ===

ID		LEADER
160cd6c5	false
4b7963aa	false
ee965815	true

=== Active proxies ===

ID
8ab614aa
af07a53f
b4e28181

=== Keepers ===

UID	HEALTHY	PG LISTENADDRESS	PG HEALTHY	PG WANTEDGENERATION	PG CURRENTGENERATION
keeper0	true	10.4.2.218:5432		false		3			3
keeper1	true	10.4.0.48:5432		true		2			2
keeper2	true	10.4.1.42:5432		true		5			5

=== Cluster Info ===

Master: keeper2

===== Keepers/DB tree =====

keeper2 (master)
├─keeper1
└─keeper0

pedromvieira:
After digging logs with "kubectl logs -f stolon-keeper-0", I founded the issue and a keeperremove solved the issue.

`stolonctl removekeeper keeper0 --cluster-name kube-stolon --store-backend kubernetes --kube-resource-kind=configmap`

mmpei:
so the slave starts to resync from master after you remove it from cluster?
i want to know how to add a slave to cluster or how to do when a slave becomes inconsistent with master in k8s.
create a new PV then increase the replica of keeper, then removekeeper?
@pedromvieira @sajal

Issue #562, opened 2018-09-08
Keeper can't rejoin cluster on AWS storage optimized instances


merlineus:
Hi,

We are investigating Stolon usage in our projects and discovered that keeper can't rejoin cluster after instance type changing on AWS storage optimized nodes (i3, c5d). This operation leads to keeper's storage lost due to ephemeral nature of attached SSD disks in AWS. Of course we changing instance type one by one so cluster is healthy at any time.

Logs:
```
2018-09-06T08:45:59.019Z	INFO	cmd/keeper.go:1049	current db UID different than cluster data db UID	{"db": "", "cdDB": "b408a597"}
2018-09-06T08:45:59.019Z	ERROR	cmd/keeper.go:1346	different local dbUID but init mode is none, this shouldn't happen. Something bad happened to the keeper data. Check that keeper data is on a persistent volume and that the keeper state files weren't removed
```

Could you please suggest the right way to handle this case with Stolon?

We are using Stolon v0.12.0.

sgotti:
@merlineus the keeper needs persistent data and it'll return an error if the data has been replaced with an empty one (see the architecture doc: https://github.com/sorintlab/stolon/blob/master/doc/architecture.md#keepers).

What you should do, when replacing the data with a new one, is to firstly stop the keeper and remove it from the cluster data (`stolonctl removekeeper`), then restart it with the new data.

sgotti:
@merlineus I hope I answered your question. If you have further questions please (if it's not a possible bug/enhancement) discuss this on the gitter channel or the mailing list: https://github.com/sorintlab/stolon#contacts

merlineus:
Yes, thank you!

PR #738, opened 2019-12-03
add docs for manually resyncing a broken replica

satyanash:
Given the resolutions to the following issues: #562 #604, I felt that having an explicit doc would increase visibility for the solution. Hence this PR.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -22,6 +22,7 @@ We suggest that you first read the [Stolon Architecture and Requirements](archit
 * [PostgreSQL SSL/TLS setup](ssl.md)
 * [Forcing a failover](forcefailover.md)
 * [Service Discovery](service_discovery.md)
+* [Resync a Broken Replica](resync_broken_replica.md)
 
 ### Recipes
 
--- a/doc/architecture.md
+++ b/doc/architecture.md
@@ -14,11 +14,10 @@ Stolon is composed of 3 main components
 
 #### Keepers
 
-Every keeper MUST have a different UID that can be manually provided (`--uid` option) or will be generated. After the first start the keeper id (provided or generated) is saved inside the keeper data directory.
-
-Every keeper MUST have a persistent data directory (no ephemeral volumes like k8s `emptyDir`) or you'll lose your data if all the keepers are stopped at the same time (since at restart no valid standby to failover will be available).
-
-If you're providing the keeper's uid in the command line don't start a new keeper with the same id if you're providing a different data directory (empty or populated) since you're changing data out of the stolon control causing possible data loss or strange behaviors.
+* Every keeper MUST have a different UID that can be manually provided (`--uid` option) or will be generated. After the first start the keeper id (provided or generated) is saved inside the keeper data directory. If you start two distinct keeper instances with the same UID, expect strange behaviour and/or data loss.
+* Every keeper MUST have a persistent data directory (no ephemeral volumes like k8s `emptyDir`) or you'll lose your data if all the keepers are stopped at the same time (since at restart no valid standby to failover will be available).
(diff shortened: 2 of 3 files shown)


satyanash:
@sgotti The failures look intermittent and unrelated.

doc/resync_broken_replica.md
```diff
@@ -0,0 +1,18 @@
+# Re-sync a broken replica
+
+Due to unforeseen circumstances, if you end up with an empty postgres data directory on one of your replicas but your _master_ is intact, you will receive a message that looks something like this:
+
+```
+different local dbUID but init mode is none, this shouldn't happen. Something bad happened to the keeper data. Check that keeper data is on a persistent volume and that the keeper state files weren't removed
+```
+
+This is because all keepers _require_ persistent storage and will not auto "re-sync" just because they find an empty data directory while still having an entry in the clusterdata.
```
sgotti: I'll add a link to the `architecture.md` doc that clearly explains this.

satyanash: @sgotti I've added the link under the Misc. section, and also updated `architecture.md` to specify the behaviour in case of a cluster ID mismatch. Please review.

doc/README.md
```diff
@@ -22,6 +22,7 @@ We suggest that you first read the [Stolon Architecture and Requirements](archit
 * [PostgreSQL SSL/TLS setup](ssl.md)
 * [Forcing a failover](forcefailover.md)
 * [Service Discovery](service_discovery.md)
+* [Resync a Broken Replica](resync_broken_replica.md)
```
sgotti: To keep naming consistent I'll rename it to `Reset/Resync a broken keeper`

doc/architecture.md
```diff
@@ -14,11 +14,10 @@ Stolon is composed of 3 main components
...
-
-If you're providing the keeper's uid in the command line don't start a new keeper with the same id if you're providing a different data directory (empty or populated) since you're changing data out of the stolon control causing possible data loss or strange behaviors.
+* Every keeper MUST have a different UID that can be manually provided (`--uid` option) or will be generated. After the first start the keeper id (provided or generated) is saved inside the keeper data directory. If you start two distinct keeper instances with the same UID, expect strange behaviour and/or data loss.
+* Every keeper MUST have a persistent data directory (no ephemeral volumes like k8s `emptyDir`) or you'll lose your data if all the keepers are stopped at the same time (since at restart no valid standby to failover will be available).
+* If an existing functioning keeper, finds that it's data directory is empty, it will refuse to start until the data directory is available again.
+* If an existing functioning keeper, finds that it's current Postgres Cluster ID is different from the one it previously had in the clusterdata, it will still refuse to start.
```
sgotti: This is out of scope for the current PR. If you want to change the architecture.md doc please open a new PR.

doc/resync_broken_replica.md
```diff
@@ -0,0 +1,18 @@
+# Re-sync a broken replica
```
sgotti: To keep naming consistent I'll rename it to Reset/Resync a broken keeper

doc/resync_broken_replica.md
sgotti: This is just a case of "broken keeper", there can be many other cases so I'll make it more generic and make this a possible case (empty postgres data due to **wrong** user configuration)

doc/resync_broken_replica.md
satyanash: @sgotti I use the term "replica" because one cannot resync _any_ keeper. The master has to be intact for this to work.

sgotti: What about `Reset/Resync a broken standby keeper`?

satyanash: Will incorporate this suggestion.

Issue #678, opened 2019-06-26
`check function error` when `enableServiceLinks` option set in k8s-1.14.3


icefed:
### Submission type

 - Bug report

### Environment
Kubernetes 1.14.3
#### Stolon version
v0.13.0-pg9.6
#### Additional environment information if useful to understand the bug
Add `enableServiceLinks: false` option to stolon-* PodSpec, Keeper and Proxy print error and can not work.

### Unexpected behaviour you saw

Stolon-Proxy error output

[...]

PR #740, opened 2019-12-07
Make the k8s store properly patch annotations

sandhose:
Hi,

While trying out Stolon on Kubernetes, I ran into a weird issue where Stolon was unable to change the annotations on the sentinel and keeper pods. The api-server said it was trying to change some fields that were immutable. This was due to Stolon trying to replace the whole pod object instead of patching it.

This PRs changes the k8s store to properly use jsonpatch to set the annotations. This is heavily inspired by the way `kubectl annotate` does patch the annotations: https://github.com/kubernetes/kubectl/blob/b909fcb4a071a1a9669a9fe1f48482c848823124/pkg/cmd/annotate/annotate.go#L277-L311

### DIFF
--- a/internal/store/k8s.go
+++ b/internal/store/k8s.go
@@ -23,10 +23,12 @@ import (
 	"github.com/sorintlab/stolon/internal/cluster"
 	"github.com/sorintlab/stolon/internal/util"
 
+	jsonpatch "github.com/evanphx/json-patch"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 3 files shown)


dineshba:
Code looks good to me. @sandhose How to test this feature?

sgotti:
@sandhose Thanks for the PR. Issue #678 is probably related. I'll review it in the next days.

ccmehk:
This look good, i want a quick patch on this

internal/store/k8s.go
```diff
@@ -216,11 +219,35 @@ func (s *KubeStore) SetKeeperInfo(ctx context.Context, id string, ms *cluster.Ke
...
+		if err != nil {
+			return err
+		}
+
+		annotations := accessor.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		annotations[util.KubeStatusAnnnotation] = string(msj)
+		accessor.SetAnnotations(annotations)
```
sgotti: I don't understand why this is needed, isn't this doing the same thing at the previous single line?

```
result.Annotations[util.KubeStatusAnnnotation] = string(msj)
```

If so all of this can be summarized to:

```
		oldData, err := json.Marshal(result)
		if err != nil {
			return err
		}

		result.Annotations[util.KubeStatusAnnnotation] = string(msj)

		newData, err := json.Marshal(result)
		if err != nil {
			return err
		}

		patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
		if err != nil {
			return err
		}

		_, err = podsClient.Patch(s.podName, types.MergePatchType, patchBytes)
		return err
```

sandhose: You're totally right. That's because I borrowed the code from `kubectl annotate`, which is designed to be a lot more generic, hence the accessor. I'll fix it.

sgotti:
@sandhose let me know when you're ready for another review (I saw you pushed an updated but it didn't fixed my comments).

If you're interested this is what I had in mind to simplify this PR: https://github.com/sgotti/stolon/commit/de8eddd69337989ba7b63587179b844e8a7cd3f6

Please also update the commit message to follow our conventions and better describe what this patch does:

```
k8s store: patch pod annotations

Use k8s client mpatch method to update pod annotations (in some versions
replacing the whole pod object will return an error)
```

Josua-SR:
@sandhose I came across a similar issue when I was trying to deploy stolon, and this PR definitely solved it on my end - thank you very much!

sgotti:
@sandhose Since there was no activity and I'd like to merge this soon I created a new PR #751 with the proposed changes.

sandhose:
@sgotti Thanks for picking that up, I really did not had time to finish this those past weeks

sgotti:
@sandhose Thanks to you! I left you as the author of the commit in the new PR.

PR #741, opened 2019-12-09
k8s e.g.: Enable high availablility by NOT collocating pods on same node

Miyurz:
With this change, we ensure that the replica pods of the same
deployment do not get assigned to the same node. Fot this, we have
employed k8s anti-affinity rule.


mayurgn86:
@sgotti Can you please review ?

Issue #633, opened 2019-04-12
add ability to defice custom entries in pg_ident.conf


dbarkin:
### Submission type

 - Request for enhancement (RFE)

### Environment
Linux Ubuntu/Redhat
#### Stolon version
v0.13.0
#### Additional environment information if useful to understand the bug
we use peer authentication and have scripts which require mapping to be defined in pg_ident.conf

### Expected behaviour you didn't see
ability to use stolonctl to update pg_ident.conf on all keepers
### Unexpected behaviour you saw
now we have to use manual edit on every keeper
### Steps to reproduce the problem
try to update pg_ident.conf

### Enhancement Description
Please add new functionality to add/update custom entries pg_ident.conf with stolonctl in similar way how it is done with custom pgHBA.

dineshba:
@sgotti Is this feature needed? If so, Can I work on it?

sgotti:
@dbarkin every contribution is always welcome! I haven't digged enough on how pg_ident.conf works so I can't give you much guidance. I hope it could be implemented in the save way we implemented the handling of pg_hba.conf

dineshba:
Sure. I will refer the implementation of `pg_hba.conf` and try to implement in the same way. Thanks

mpniel:
Hello
Is there any news about the adding entries to pg_ident.conf ?

PR #753, opened 2020-01-22
[WIP] add ability to define custom entries in pg_ident.conf

dineshba:
Trying to fix https://github.com/sorintlab/stolon/issues/633

Tasks:
- [x] Able to give pg_ident content in clusterSpec
- [x] Update `pg_ident.conf` file whenever there is a change in pg_ident in clusterSpec
- [ ] Make changes in `pg_hba.conf` as per [here](https://www.postgresql.org/docs/9.1/auth-username-maps.html)

#### How to use in clusterSpec:
```json
stolonctl update -p '{"pgIdent": {"omicron": [{"systemUsername": "bryanh", "databaseUsername": "bryanh"}, {"systemUsername": "robert", "databaseUsername": "bob"}]}}'
```

> Note: Raising WIP PR to get faster feedback on the approach.


sgotti:
@dineshba Thanks for the PR!

* Is it ready for review?
* Do you think we need some integration tests for this?
* We should probably document the new parameter in the cluster spec.

dineshba:
Hi @sgotti 

It is better to add some integration tests. So I will add `integration test` and update the document about the new parameter. Thanks for the suggestion

So, it is not ready for review. I will mark the PR as WIP.

internal/postgresql/postgresql.go
```diff
@@ -877,6 +907,25 @@ func (p *Manager) writePgHba() error {
...
 }
 
+func (p *Manager) writePgIdent() error {
+	return common.WriteFileAtomicFunc(filepath.Join(p.dataDir, "pg_ident.conf"), 0600,
+		func(f io.Writer) error {
+			if p.ident != nil && len(p.ident) > 0 {
+				if _, err := f.Write([]byte("# MAPNAME\tSYSTEM-USERNAME\tPG-USERNAME" + "\n")); err != nil {
+					return err
+				}
+				for key, value := range p.ident {
```
porty, line 917: Entries will likely be written out in an undefined order (https://go.dev/blog/maps#iteration-order) - sorting the entries before writing them out might be beneficial

internal/postgresql/postgresql.go
```diff
@@ -877,6 +907,25 @@ func (p *Manager) writePgHba() error {
...
+					for _, v := range value {
+						if _, err := f.Write([]byte(fmt.Sprintf("%s\t%s\t%s", key, v.SystemUsername, v.DBUsername) + "\n")); err != nil {
```
porty, line 919: `f` is an `os.File` which doesn't guarantee full writes

porty:
I found that attempting to replace an entry for one of the ident maps fails due to `apimachinery` unable to merge map entries (or something along those lines). Attempting to do so gives this error:

```
$ bin/stolonctl update \
    --cluster-name stolon-cluster \
    --store-backend=etcdv3 \
    -p '{ "pgIdent": { "map1": [{ "systemUsername": "sysuser1", "databaseUsername": "dbuser1" }, { "systemUsername": "sysuser2", "databaseUsername": "dbuser2" }] } }'
failed to patch cluster spec: failed to merge patch cluster spec: merging an object in json but data type is not struct, instead is: map
```

Overall I think structuring the `pg_ident.conf` entries in to a nice JSON representation, only to put them back as lines in a file is overkill, and confusing - I've found that using `[]string` instead (similar to pgHBA) is simpler and less error prone.

PR #764, opened 2020-02-26
*: update to go 1.14

sgotti:
* Update to go 1.14
* Update golangci-lint to v1.23.6 that works with go 1.14


sgotti:
We'll wait on https://github.com/golang/go/issues/37436

PR #791, opened 2020-07-14
add keeper flag to generate pgHBA config on restarts

FarhanSajid1:
Changes:
* added in additional flag to control whether we want to skip the rendering of the hba configuration file defaulting to true to preserve all existing behavior

Reasons:
Currently in our setup, there is synchronous replication which means when the cluster is brought down [stolon doesn't regenerate](https://github.com/nsone/stolon/blob/master/cmd/keeper/cmd/keeper.go#L1482-L1488) the custom `pgHBA` files from the `postgresql.json` file. This could potentially lead to missing auth entries in the `pg_hba.conf` file, and lead to a situation where the cluster never comes up healthy 
Example:
`no pg_hba.conf entry for host \"127.0.0.1\", user \"ns1\", database \"ns1\", SSL off`

There should be a flag to trigger the generation of this file in this situation to on/off, but please let us know if that doesn't make sense or there's a flaw in that logic that would result in unwanted behavior.

### DIFF
--- a/cmd/keeper/cmd/keeper.go
+++ b/cmd/keeper/cmd/keeper.go
@@ -143,7 +143,7 @@ func init() {
 
 	CmdKeeper.PersistentFlags().BoolVar(&cfg.canBeMaster, "can-be-master", true, "prevent keeper from being elected as master")
 	CmdKeeper.PersistentFlags().BoolVar(&cfg.canBeSynchronousReplica, "can-be-synchronous-replica", true, "prevent keeper from being chosen as synchronous replica")
-	CmdKeeper.PersistentFlags().BoolVar(&cfg.debug, "skip-render-hba", true, "boolean to render the hba configuration file even on restarts")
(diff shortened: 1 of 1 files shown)


sgotti:
@FarhanSajid1 If you're encountering an issue please open a dedicated issue with a way to reproduce it (or better a test case) before opening a PR so we'll be able to understand and find a fix. I don't think that there's the need to add a flag to fix it.

pgray:
@sgotti we've recently had some users fully stop (all at once) a cluster of 3 nodes.
on startup (sometimes), the cluster comes up, goes healthy, but then the old master is re-elected and
this setting prevents it from becoming the cluster master (or at least blocks our postgres connection into it).
A subsequent kick fixes it and it re-renders the config.

FarhanSajid1:
> @FarhanSajid1 If you're encountering an issue please open a dedicated issue with a way to reproduce it (or better a test case) before opening a PR so we'll be able to understand and find a fix. I don't think that there's the need to add a flag to fix it.

Related issue: https://github.com/sorintlab/stolon/issues/792

sgotti:
Closing since this isn't really needed.

Issue #793, opened 2020-07-22
Hide root & replication users password logs from Postgres


keierk01:
**What would you like to be added**:
Hide `root` and `replication` roles passwords from Postgres logs in startup phase

**Why is this needed**:
Even logs are usually treated as confidential content it's not good idea to have any passwords visible there. Currently Stolon use `CREATE ROLE` or `ALTER ROLE` commands to set password for Postgres roles. This generates Postgres log entry as cleartext password included. See example below.

```
2020-07-10 15:16:53 UTC:[local]:postgres@postgres:[67]:LOG:  statement: alter role postgres with password 'PostgresRootPassword';
2020-07-10T15:16:53.315Z        INFO    postgresql/postgresql.go:538    superuser password set
2020-07-10T15:16:53.315Z        INFO    postgresql/postgresql.go:541    creating replication role
2020-07-10 15:16:53 UTC:[local]:postgres@postgres:[68]:LOG:  statement: create role "stolon_repluser" with login replication encrypted password 'StolonReplUserPassword';
2020-07-10T15:16:53.319Z        INFO    postgresql/postgresql.go:551    replication role created        {"role": "stolon_repluser"}
```

Possible solution would be to hide Postgres logs temporary when altering or creating Postgres role. See example from [stackexchange](https://dba.stackexchange.com/questions/94625/how-do-i-hide-sensitive-information-like-plaintext-passwords-from-the-logs)

```
BEGIN;
SET LOCAL log_statement = 'none';
ALTER USER ... SET PASSWORD ...;
COMMIT;
```

Something like below with Go when set password to Postgres role.
```
func setPassword(ctx context.Context, connParams ConnParams, username, password string) error {
	db, err := sql.Open("postgres", connParams.ConnString())
	if err != nil {
		return err
	}
	defer db.Close()

	tx, err := db.Begin()
	if err != nil {
		return err
	}

	// Disable logs temporary to prevent password expose in logs
	query := fmt.Sprintf("SET LOCAL log_statement = %s", pg.QuoteLiteral("none"))
	if _, err := tx.Exec(query); err != nil {
		err := tx.Rollback()
		return err
	}

	query = fmt.Sprintf("ALTER ROLE %s WITH PASSWORD %s", pg.QuoteIdentifier(username), pg.QuoteLiteral(password))
	if _, err := tx.Exec(query); err != nil {
		err := tx.Rollback()
		return err
	}
	tx.Commit()

	return err
}
```

[...]

PR #795, opened 2020-08-20
Issue #793 - Hide root & replication users password from Postgres logs

keierk01:
* This change disable Postgres to print out password to logs even log_statement is something else than none (default)

Below is explained what happens when user is created or altered
1. Transaction is created when starting to manipulate password for the user
2. Logs are disabled by setting `log_statement` to `none` during the transaction

Because logs are disable during the transaction, passwords are not visible in logs at all.

### DIFF
--- a/internal/postgresql/utils.go
+++ b/internal/postgresql/utils.go
@@ -68,8 +68,23 @@ func setPassword(ctx context.Context, connParams ConnParams, username, password 
 	}
 	defer db.Close()
 
-	_, err = dbExec(ctx, db, fmt.Sprintf(`alter role "%s" with password '%s';`, username, password))
-	return err
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+
+	query := fmt.Sprintf("set local log_statement = %s", pq.QuoteLiteral("none"))
+	if _, err = tx.ExecContext(ctx, query); err != nil {
+		_ = tx.Rollback()
+		return err
+	}
+
+	query = fmt.Sprintf("alter role %s with encrypted password %s", pq.QuoteIdentifier(username), pq.QuoteLiteral(password))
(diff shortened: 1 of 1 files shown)


keierk01:
Could somebody restart the failing tests. I can't do that by myself. Thanks in advance.

keierk01:
@sgotti It looks that there are some problems in your build system, because all recently pushed PR's are failing. Including this one. Can you help me to get this proceeding?

Thanks!

internal/postgresql/utils.go
```diff
@@ -68,8 +68,23 @@ func setPassword(ctx context.Context, connParams ConnParams, username, password
...
-	_, err = dbExec(ctx, db, fmt.Sprintf(`alter role "%s" with password '%s';`, username, password))
-	return err
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+
+	query := fmt.Sprintf("set local log_statement = %s", pq.QuoteLiteral("none"))
+	if _, err = tx.ExecContext(ctx, query); err != nil {
+		err = tx.Rollback()
```
sgotti, line 78: assigning tx.Rollback to error will override the real error. I'll just ignore the Rollback error.

sgotti (changes requested):
@keierk01 Sorry but I lost your PR. Some comments inline.

The build system is working. There're just some flakes due to big resources required by the tests that are running in parallel to speed them up. Nothing to worry about.

internal/postgresql/utils.go
keierk01: @sgotti That's true. I'm not so familiar with Go, but I agree with you. I updated the PR already.

sgotti:
@keierk01 Thanks a lot! Merging.

luomengY:
When will v0.18.0 be released?

PR #804, opened 2020-10-16
Defer tabOut.Flush

edwinakatosh:
When running stolonctl status, i'm seeing the following output:

```
UID     HEALTHY PG LISTENADDRESS        PG HEALTHY      PG WANTEDGENERATION     PG CURRENTGENERATION
keeper0 true    10.1.0.36:5432          true            4                       4
keeper1 true    10.1.0.47:5432  true    2       2
```

instead of

```
UID     HEALTHY PG LISTENADDRESS        PG HEALTHY      PG WANTEDGENERATION     PG CURRENTGENERATION
keeper0 true    10.1.0.36:5432          true            2                       2
keeper1 true    10.1.0.47:5432          true            2                       2
```

Example: [https://play.golang.org/p/K9kBnGbYBEE](https://play.golang.org/p/K9kBnGbYBEE)

### DIFF
--- a/cmd/stolonctl/cmd/status.go
+++ b/cmd/stolonctl/cmd/status.go
@@ -123,8 +123,8 @@ func renderText(status Status, generateErr error) {
 	} else {
 		fmt.Fprintf(tabOut, "ID\tLEADER\n")
 		for _, s := range status.Sentinels {
+			defer tabOut.Flush()
 			fmt.Fprintf(tabOut, "%s\t%t\n", s.UID, s.Leader)
-			tabOut.Flush()
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


edwinakatosh:
@sgotti Please review and advise.

sgotti (changes requested):
@edwinakatosh Thanks for the PR and sorry for being late.
We could clean it up without doing multiple defer but just doing a `tabOut.Flush` at the end of all the writing to `tabOut` in renderText and also remove the unused tabOut defined [here](https://github.com/sorintlab/stolon/pull/804/files#diff-c24e4cb37a836828f31c4292ea8545403987c9ef416e37c628cd12e49bcc1eafL236-L237).

edwinakatosh:
I think i had tested that, but i'll confirm and update. Thanks @sgotti

PR #817, opened 2021-01-05
Add flag skip-lock-file to keeper command

alessandro-sorint:
Add flag skip-lock-file to keeper command
We use different volumes for any stolon keeper, so the lock file is not needed


cmd/keeper/cmd/keeper.go
```diff
@@ -142,6 +143,7 @@ func init() {
 
 	CmdKeeper.PersistentFlags().BoolVar(&cfg.canBeMaster, "can-be-master", true, "prevent keeper from being elected as master")
 	CmdKeeper.PersistentFlags().BoolVar(&cfg.canBeSynchronousReplica, "can-be-synchronous-replica", true, "prevent keeper from being chosen as synchronous replica")
+	CmdKeeper.PersistentFlags().BoolVar(&cfg.skipLockFile, "skip-lock-file", false, "prevent keeper to skip lock file. Warning! It'll cause data corruptions if two keepers are concurrently running with the same data dir.")
```
sgotti, line 146: I'll propose to call it `disable-data-dir-locking` since this is the behavior and in future how it's implemented may change.

Also `prevent keeper to skip lock file` is quite confusing, What about something like "disable locking on data dir. Warning .... "

sgotti (changes requested):
@alessandro-sorint Thanks for the PR. Just a small nit on flag name and description.

cmd/keeper/cmd/keeper.go
alessandro-sorint: @sgotti  Thanks for the suggestion I did the push to the pull request with "disable-data-dir-locking"

sgotti (changes requested):
@alessandro-sorint Thank! It LGTM. Just one last thing before merging it: can you please rebase against the current master?

alessandro-sorint:
@sgotti  I rebased to the current master thanks

sgotti (approved):

sgotti:
@alessandro-sorint Thanks! Merging.

Issue #794, opened 2020-07-31
sentinel don't failover on node fail


deepdivenow:
**What happened**:
This problem is not exists on stolon versions 0.10.0 or 0.15.0

Environment:
3 centos v7 or v8 nodes
each node installed:
- etcd 3.3.9 / 3.4.6 /3.4.10   API:v3
- stolon sentinel / proxy / keeper 0.16.0
- postgresql 10

[root@centos17 ~]# stolonctl status
=== Active sentinels ===

ID		LEADER
718f47de	false
79a71967	true
b17ab76b	false

=== Active proxies ===

ID
5207501e
974cf4dc
b41ff2ef

=== Keepers ===

UID		HEALTHY	PG LISTENADDRESS	PG HEALTHY	PG WANTEDGENERATION	PG CURRENTGENERATION
pgcentos17	true	192.168.1.117:5432	true		5			5	
pgcentos18	true	192.168.1.118:5432	true	2	2	
pgcentos19	true	192.168.1.119:5432	true	5	5	

=== Cluster Info ===

Master Keeper: pgcentos19

===== Keepers/DB tree =====

pgcentos19 (master)
├─pgcentos18
└─pgcentos17

On node fail (master node)
All other sentinels holds in state:
[root@centos17 ~]# stolonctl status
=== Active sentinels ===

No active sentinels

=== Active proxies ===

No active proxies

=== Keepers ===

UID		HEALTHY	PG LISTENADDRESS	PG HEALTHY	PG WANTEDGENERATION	PG CURRENTGENERATION
pgcentos17	true	192.168.1.117:5432	true		5			5	
pgcentos18	true	192.168.1.118:5432	true	2	2	
pgcentos19	true	192.168.1.119:5432	true	5	5	

=== Cluster Info ===

Master Keeper: pgcentos19

===== Keepers/DB tree =====

pgcentos19 (master)
├─pgcentos18
└─pgcentos17

Jul 31 08:31:08 centos17 stolon-sentinel[21844]: 2020-07-31T08:31:08.797-0400        INFO        cmd/sentinel.go:1964        sentinel uid        {"uid": "79a71967"}
Jul 31 08:31:08 centos17 stolon-sentinel[21844]: 2020-07-31T08:31:08.798-0400        INFO        cmd/sentinel.go:82        Trying to acquire sentinels leadership
[...]

PR #821, opened 2021-02-17
Add ETCDv3 client more statility

itsystem:
Add ETCDv3 more statility:
- enable keepalive checks
  you must react on returned err code etcd3Lib functions
  Or enable keepalive for etcd3Lib for himself react on etcd cluster changes
  issue: https://github.com/sorintlab/stolon/issues/794
- update client libs to v.3.4.14
- grow proxy timeout to 20 sec, for skip proxy connections terminate when etcd cluster change leader
When etcd cluster changes leader node, stolon-proxy resets all connections but other components not failed.
Grow DefaultProxyTimeout 15-->20 sec can change this behaviour.

### DIFF
--- a/internal/store/kvbacked.go
+++ b/internal/store/kvbacked.go
@@ -185,6 +185,9 @@ func NewKVStore(cfg Config) (KVStore, error) {
 		config := etcdclientv3.Config{
 			Endpoints: addrs,
 			TLS:       tlsConfig,
+			DialTimeout: time.Second,
+			DialKeepAliveTime:    time.Second,
+			DialKeepAliveTimeout: 500 * time.Millisecond,
 		}
 
 		c, err := etcdclientv3.New(config)
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/sorintlab/pollon v0.0.0-20181009091703-248c68238c16
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
-	go.etcd.io/etcd v0.0.0-20191023171146-3cf2f69b5738
+	go.etcd.io/etcd v0.5.0-alpha.5.0.20201125193152-8a03d2e9614b
 	go.uber.org/zap v1.13.0
 	k8s.io/api v0.17.3
 	k8s.io/apimachinery v0.17.3
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -56,7 +56,7 @@ const (
 	DefaultFailInterval                               = 20 * time.Second
 	DefaultDeadKeeperRemovalInterval                  = 48 * time.Hour
 	DefaultProxyCheckInterval                         = 5 * time.Second
-	DefaultProxyTimeout                               = 15 * time.Second
+	DefaultProxyTimeout                               = 20 * time.Second
 	DefaultMaxStandbys               uint16           = 20
 	DefaultMaxStandbysPerSender      uint16           = 3
 	DefaultMaxStandbyLag                              = 1024 * 1204


itsystem:
I think tests have failed not by code reasons
Successfully tagged stolon:master-pg11
/stolon/examples/kubernetes /stolon
error: unable to recognize "role.yaml": Get https://localhost:8443/api?timeout=32s: dial tcp 127.0.0.1:8443: connect: connection refused

internal/store/kvbacked.go
```diff
@@ -185,6 +185,9 @@ func NewKVStore(cfg Config) (KVStore, error) {
 		config := etcdclientv3.Config{
 			Endpoints: addrs,
 			TLS:       tlsConfig,
+			DialTimeout: time.Second,
```
sgotti, line 188: DialTimeout is just related to creating a connection to an endpoint (the it's kept open), making it of just 1 second is probably too much (k8s uses 20 seconds) and I'm not sure which issue will fix.

internal/store/kvbacked.go
```diff
@@ -185,6 +185,9 @@ func NewKVStore(cfg Config) (KVStore, error) {
...
+			DialKeepAliveTime:    time.Second,
+			DialKeepAliveTimeout: 500 * time.Millisecond,
```
sgotti, line 190: Also these are too small. What about making the timeout to the current context timeout of 5 seconds?

sgotti (changes requested):
@itsystem thanks for your PR!

Can you provide more insights on what issue this change will fix?

In our enviroments everything works correctly also without it and when an etcd node dies we failover to another node. During some tests I noticed that if the tcp connection isn't reset than we had some issue that setting KeepAliveTimeout will fix. To reproduce this locally I just SIGSTOP one etcd process. So adding a KeepAliveTimeout is definitely a good thing to do.

> grow proxy timeout to 20 sec, for skip proxy connections terminate when etcd cluster change leader
When etcd cluster changes leader node, stolon-proxy resets all connections but other components not failed.
Grow DefaultProxyTimeout 15-->20 sec can change this behaviour.

Are you saying that etcd takes too much time to elect a new leader in relation to the stolon proxy timeout and you want to avoid the timeout right? If so, is the etcd leader election time something depending on multiple environment conditions or something with a defined upper bound? We should understand if increasing the timeout works only on your enviroment or also everywhere (and probably make the proxy timeout configurable, there's already an open issue for this).

Some comments on the patch are inline.

deepdivenow:
Hi I have writed video's for you

This video more short and you can see how two sentinel versions work in one situation when master node is disappeared
https://youtu.be/4xto0y1EfHw

This more long video only v0.16.0 in this video all stolon components in failed state before i restarted theirs.
https://youtu.be/HJZbeR2BGFs

internal/store/kvbacked.go
deepdivenow: I think this timeout not matter as keepalive

internal/store/kvbacked.go
deepdivenow: I thing you can grow it as you think be the best

deepdivenow:
> Are you saying that etcd takes too much time to elect a new leader in relation to the stolon proxy timeout and you want to avoid the timeout right? If so, is the etcd leader election time something depending on multiple environment conditions or something with a defined upper bound? We should understand if increasing the timeout works only on your enviroment or also everywhere (and probably make the proxy timeout configurable, there's already an open issue for this).

If this timeout be configurable is better because leader election time is configurable in etcd cluster.
And fixed timeout can't satisfy all env.

sgotti:
@deepdivenow I created two PRs in place of this: #827 #828

> If this timeout be configurable is better because leader election time is configurable in etcd cluster.
> And fixed timeout can't satisfy all env.

This was already implemented in #756

Thanks!

PR #838, opened 2021-07-14
Migrate etcd to use 'etcd-io/etcd' and update relevant paths

davissp14:
Given the `coreos/etcd` client was deprecated back in 2015, let's work to move to `etcd-io/etcd`.


davissp14:
`make tests` in Agola are failing with:

```
panic: codecgen version mismatch: current: 8, need 10. Re-generate file: /go/pkg/mod/github.com/coreos/etcd@v3.3.10+incompatible/client/keys.generated.go
goroutine 1 [running]:
```
Not quite sure what to do about that.  Tests work fine locally.

sgotti:
@davissp14  Thanks for the PR. Sorry but I don't understand the title and why this is needed. We're already using for etcdv3 go.etcd.io/etcd client v3, just the 3.4 version with a workaround since only from the recently released etcd v3.5 they fixed the wrong module path (missing /v3). 
`github.com/coreos/etcd` is used for etcdv2 and required by libkv.

davissp14:
@sgotti My mistake, the dependencies here are a little confusing...  We can discuss this elsewhere if you'd like.   I have a fork where i'm introducing the ability to specify Store specific credentials for Etcd.  I was not able to get authentication to work properly using the currently client version.  Bumping the client to 3.5 seems to address my issues.

PR #843, opened 2021-09-09
Allow MinSyncRepl = 0 (from #841)

rg2011:
Comes from PR #841. Squishes the following changes in a single commit:

* Added support for "strictSyncRepl" cluster parameter
* Fixed keeper to support empty syn replica list
* Added integration tests
* Fixed missing return value check
* Removed noisy warning
* check tests fail because of load
* revert test to original setup
* Added more HA tests
* removed StrictSyncRepl, added support for MinSyncStandby == 0
* Checking MinSynchronousStandbys for < 0 fails
* Removed integration ha_tests not affected by MinSyncStandbys == 0
* Removed uneeded check
* Removed unneeded sync0 test
* fixed doc typo
* Update tests/integration/ha_test.go


sgotti:
@rg2011 Instead of opening a new PR you could just squash (using git rebase) and then force push the branch.

sgotti:
@rg2011 Closing this so we can continue in the original PR #841

PR #848, opened 2021-10-05
Add hostname's to stolonctl output

deepdivenow:
Add stolon components hostnames to stolonctl output.
`# stolonctl status
=== Active sentinels ===

ID		LEADER	HOSTNAME
06d77408	false	centos19
0a2ab67c	false	centos18
58b4fab7	true	centos17

=== Active proxies ===

ID		HOSTNAME
4332e2ae	centos18
71a18a24	centos17
b157e411	centos19
b5b6326f	centos17
c38c5d6e	centos18
e0d908fc	centos19

=== Keepers ===

UID		HEALTHY	PG LISTENADDRESS	PG HEALTHY	PG WANTEDGENERATION	PG CURRENTGENERATION
pgcentos17	true	10.10.1.117:5432	true		3			3	
pgcentos18	true	10.10.1.118:5432	true	2	2	
pgcentos19	true	10.10.1.119:5432	true	2	2	

=== Cluster Info ===

Master Keeper: pgcentos17
`


internal/common/common.go
```diff
@@ -137,3 +138,19 @@ func WriteFileAtomic(filename string, perm os.FileMode, data []byte) error {
...
+func GetHostname() string {
+	once.Do(func() {
+		var err error
+		hostname, err = os.Hostname()
+		if err != nil {
+			hostname = ""
+		}
+	})
+	return hostname
+}
```
sgotti, line 156: I'll put this in a dedicated global struct instead of having two global variables with very generic (`once`) names.

internal/common/common.go
```diff
@@ -140,17 +140,21 @@ func WriteFileAtomic(filename string, perm os.FileMode, data []byte) error {
...
 		var err error
-		hostname, err = os.Hostname()
+		h.hostname, err = os.Hostname()
 		if err != nil {
-			hostname = ""
+			h.hostname = ""
 		}
 	})
-	return hostname
+	return h.hostname
```
sgotti: Sorry, I wasn't clear. I mean something like:

``` go
var (
	h struct {
		once     sync.Once
		hostname string
	}
)

func GetHostname() string {
	h.once.Do(func() {
		log.Printf("hostname get called")
		var err error
		h.hostname, err = os.Hostname()
		if err != nil {
			h.hostname = ""
		}
	})
	return h.hostname
}


```
where the struct is totally hidden and you just call `common.GetHostname()`.

Anyway I don't think that calling os.Hostname is a big overhead so we should also avoid the sync once.

sgotti (changes requested):

internal/common/common.go
deepdivenow: Maybe Get os.Hostname once in init section.
Like




`
func init(){
	setHostname()
}
var (
	hostname string
)

func GetHostname() string {
	return hostname
}

func setHostname() {
	var err error
	hostname, err = os.Hostname()
	if err != nil {
		hostname = ""
	}
}
`
And return global var hostname from func GetHostname.

sgotti: Seems good.

deepdivenow: Update to one clear commit

deepdivenow:
@sgotti I want firstly see names for proxy&sentinels. (i always know where my keepers)
I create this AS IS because sentinel change info about keepers only if is new keeper.
I try add this info in CD but hostname shown only for new keepers, for backward compatibility previous version of stolon, i create AS IS.

PR #865, opened 2022-02-18
support --wal-dir in keeper

sebasmannem:
Trying to get the --wal-dir option from #641 merged.


internal/postgresql/postgresql.go
```diff
@@ -299,10 +299,98 @@ func (p *Manager) StartTmpMerged() error {
...
+	log.Debugf("moving WAL files from %s to %s", curPath, tmpPath)
+	if entries, err := ioutil.ReadDir(curPath); err != nil {
+		log.Errorf("could not read contents of folder %s: %e", curPath, err)
+		return err
+	} else {
+		for _, entry := range entries {
+			srcEntry := filepath.Join(curPath, entry.Name())
+			dstEntry := filepath.Join(tmpPath, entry.Name())
+			log.Debugf("moving %s to %s", srcEntry, dstEntry)
+			if err = os.Rename(srcEntry, dstEntry); err != nil {
```
VSpike, line 344: This method will only work with a single filesystem. Cross filesystem moves will throw an error. This is probably not idea since one of the main uses of this flag would be to put the WAL data on a separate volume.

sebasmannem: Ouch. I was testing in a docker container with one MP.
Good catch. I will submit a fix.

internal/postgresql/postgresql.go
```diff
@@ -299,10 +299,98 @@ func (p *Manager) StartTmpMerged() error {
...
+		}
+	} else {
+		err := fmt.Errorf("location %s is no symlink and no dir, so please check and resolve by hand", symlinkPath)
+		log.Error(err)
+		return err
+	}
+	if p.walDir == "" {
+		// So we were moving WAL files back into PGDATA. Let's rename the tmpDir now holding all WAL files and use that
+		// as PGDATA/pg_wal
+		if err = os.Rename(tmpPath, desiredPath); err != nil {
```
VSpike, line 357: Same applies here! This won't work cross-filesystem.

sebasmannem:
@sgotti this is ready to be merged.
Can you have a look?

sgotti:
@sebasmannem Thanks. I'll take a look at this in the next days.

In the meantime there're some lint errors on go1.16. For go 1.15 you're using a package not available in such version but I'll bump the go versions to the last two major versions in another PR.

sebasmannem:
Thanks 🙏 

Verstuurd vanaf mijn iPhone

sebasmannem:
@sgotti I intend to finalise this and my other commit over the next few weeks.
How is your bump PR going? Anything I can do to help?

PR #868, opened 2022-03-10
Get tests working for local dev

dyson:
Get tests working for local development - primarily around only running
gofmt if locally you are using go v1.12.

In CI we will build with go v1.12 so won't have to worry about this.


dyson:
Sorry for the noise wrong repo!

Issue #869, opened 2022-03-16
Have stolon work with client certs


sebasmannem:
Hi,
I want to use stolon with client certs.
That would be other clients, but stolon as well.
I noticed that --pg-su-auth-method, and --pg-repl-auth-method don't support a cert value, that is easilly fixed.
But I also noticed that the first columns for their hba lines are always host and they shoulb be hostssl Making them configurable too is another small and easy addition.
After that I think it will work with using client certs for stolon.

**What would you like to be added**:
The option for stolon to communicate with client certificates.

**Why is this needed**:
For high security environments this is a much appreciated capability.

sebasmannem:
I will do some preliminary tests and issue a PR.
Just wanted all to know that I will be working on this.

sebasmannem:
Using this for 3 years without any issues.
I have releases on github.com/pgvillage-tools/stolon that have this (and other things).

PR #870, opened 2022-03-16
enabling client certs

sebasmannem:
Trying to have stolon work with client certificates


sebasmannem:
Fixes #869

sebasmannem:
@sgotti this now works. It adds a lot of options, like peer, ident and cert auth.
It also adds an option to set hostssl instead of host.
This implementation is very configurable, and we could add documentation how to deploy a client cert authorised setup.
Alternatively we could wrap all in one or 2 options which would be easier to setup and document, but would not be as flexible.
Let me know your opinion please...

PR #890, opened 2022-09-20
feat: add retry times when request store

shangjin92:
Sometimes, the storage is not stable, and cannot get cluster data from storage, or maybe timeout. It will cause to stop listening, and disconnect current connection, but the keeper is actually healthy. So add retry method when request cluster data, it will make system more reliable.


sgotti:
@shangjin92 This will defeat the purpose of the stolon proxy: stop connections when we can't know the cluster state. There's already a retry done by TimeoutChecker until ProxyTimeout. See https://github.com/sorintlab/stolon/blob/master/doc/architecture.md

PR #896, opened 2022-10-17
*: add support for PostgreSQL 15

chhetripradeep:
Now that [Postgres 15](https://www.postgresql.org/docs/15/release-15.html) is out, it will be good to support it.


sgotti:
@chhetripradeep Thanks for the PR. The CI required a dedicated test image for PostgreSQL 15 that I pushed. Beside some CI glitches look like no changes are needed for Postgres 15. Have you tried it in the real life?

Issue #888, opened 2022-09-12
Proxy does not recover after "Stopping listening"


nh2:
**What happened**:

There was a temporary Consul downtime in my cluster. After `Stopping listening`, the `stolon-proxy` does not recover:

```
2022-09-11T22:34:54.630Z        INFO        cmd/proxy.go:304        check timeout timer 
2022-09-11T22:34:54.630Z        INFO        cmd/proxy.go:158        Stopping listening
```

No more output for multiple hours, even though Consul evidently recovered (Consul checks on the same machine succeeded).

**What you expected to happen**:

That after `Stopping listening` it eventually starts listening again, as written on https://github.com/sorintlab/stolon/pull/76 it says:

> It will restart listening on successful etcd communication and a populated proxyConf.

**How to reproduce it (as minimally and precisely as possible)**:

Not sure yet, I observed it after a temporary network hickup that temporarily made Consul connections fail.

**Anything else we need to know?**:

**Environment**:
- Stolon version: `0.16.0`
- Stolon running environment (if useful to understand the bug): NixOS Linux

nh2:
I worked around it by simply restarting that `stolon-proxy`, in case that gives a hint.

nh2:
I have a suspicion of what's going on.

Reducing the above goroutine dump to the 3 goroutines that I think are relevant:

```go
goroutine 1 [select, 16591 minutes]:
github.com/sorintlab/stolon/cmd/proxy/cmd.(*ClusterChecker).Start(0xc00003c500, 0x8, 0x0)
        github.com/sorintlab/stolon/cmd/proxy/cmd/proxy.go:338 +0x171
	...

goroutine 7 [select, 16591 minutes]:
github.com/sorintlab/stolon/cmd/proxy/cmd.(*ClusterChecker).TimeoutChecker(0xc00003c500, 0xc000102180)
        github.com/sorintlab/stolon/cmd/proxy/cmd/proxy.go:302 +0xe5
created by github.com/sorintlab/stolon/cmd/proxy/cmd.(*ClusterChecker).Start
        github.com/sorintlab/stolon/cmd/proxy/cmd/proxy.go:335 +0xc5

goroutine 846743 [select, 16591 minutes]:
net/http.(*persistConn).roundTrip(0xc0004446c0, 0xc0000b1200, 0x0, 0x0, 0x0)
        net/http/transport.go:2610 +0x765
[...]

PR #907, opened 2023-04-13
proxy: Add --exit-on-store-error flag

nh2:
Workaround for #888.

`--exit-on-store-error` can be given when `stolon-proxy` is managed with something that will restart it on exit, e.g. `systemd` or similar.


nh2:
Note this workaround isn't perfect.

It will trigger the proces to exit even on temporary consul HTTP 500 errors that are recoverable.

It would be much better to distinguish between proper consul hangs, and such temporary errors.

PR #919, opened 2023-11-22
upgrade client-go version to v0.21.2, due to a bug in the v0.17.3 ver…

king-119:
…sion, when kube-api has a downtime, there will be a probability of connection timeout, and then can not select the master


ctrix:
This was integrated into MR #926 which integrates many changes and improvements.

